PR #13, opened 2023-08-07
📦Feature/maintenance update

abunur:
♻️remove redundant label
📦AGP Upgrade
📦Library updates
Update Target SDK
Updates for Gradle, Compose, Ditto SDK etc


okdistribute:
This doesn't work for me, seeing

```
Caused by: org.gradle.api.GradleException: 'compileDebugJavaWithJavac' task (current target is 11) and 'compileDebugKotlin' task (current target is 1.8) jvm target compatibility should be set to the same Java version.
```

PR #16, opened 2023-09-22
Tweak the Gradle files

anyercastillo:
### Description
The goal of the PR is to minimize our dependencies, so our artifacts do not pass those constraints to clients. Example of this problem is Ditto SDK which forces a minimum SDK to 23.

In order to achieve the goal, the following tasks were done:
- Introduces Gradle Catalog.
- Removes not used libraries from projects.
- Consolidates all android configurations in a single file.
- Sets the minSdk=23 to match Ditto's requirements.
- Aligns compile options and Kotlin's target-jvm to Java 1.8.
- Aligns compose compiler (1.4.4) and Kotlin version (1.8.10)

### An existing bug
While working on this issue I became aware of a bug that prevent our artifacts to build in Jitpack.
The follow was done to address the bug:
- Install Gradle 7.5 to fix our [compile issue](https://jitpack.io/com/github/getditto/DittoAndroidTools/0.0.1/build.log) on Jitpack

### Recommendation
Because this PR introduces a set of binaries and generated code from the Gradle install process. It is recommended you review one commit at the time.


anyercastillo:
This PR should partially fix the issue https://github.com/getditto/ditto/issues/10092

okdistribute:
Nice! Thanks for taking a look at this.

texasRanger09 (approved):

kndoshn (approved):
LGTM - Thanks!

PR #25, opened 2024-01-23
Extract DittoLogger.setLogFile(logFile.toString()) from export logs

### LINKED ISSUE #20: Bug: export logs doesn't work the first time
Unconfirmed but it was reported that the first time you touch the button to export logs, the file was empty

texasRanger09:
fixes #20 

There is a race condition happening from when you set DittoLogger.setLogFileURL and Ditto starts writing to the log file with when the file is exported. The first time you export the logs the log file is blank because the file is exported before Ditto has a chance to start writing to the log file. When we move DittoLogger.setLogFileURL out of the ExportLogs tool this gives Ditto enough time to write logs to the file before it is exported.


okdistribute (approved):

byoung-ditto:
We copy/pasted the Export Logs code into our Android project (ATAK). And made local modifications so it wont be straightforward to test this fix in ATAK. Once this is a Maven module, we can refactor ATAK to pull from there

PR #28, opened 2024-01-26
Prepare presence viewer, disk usage and export logs for deployment

texasRanger09:


DittoPresenceViewer/build.gradle
```diff
@@ -3,10 +3,14 @@ plugins {
     alias libs.plugins.org.jetbrains.kotlin.android
 }
 
+ext.libraryGroupId = "live.ditto.dittopresenceviewer"
+ext.libraryVersion = "0.0.1"
```
phatblat, line 7: We should set an explicit `artifactId` to control the maven coordinate name in case this gradle module name ever changes.
https://github.com/getditto/DittoAndroidTools/blob/69ae98ea366873a6c22f2da0517e36c5b1c1a6a7/gradle/deploy.gradle#L42C22-L44

See the [`publish.gradle`](https://github.com/getditto/ditto/blob/main/android/ditto/publish.gradle#L11) used by the Android SDK.

phatblat:
```suggestion
ext.libraryGroupId = "live.ditto"
ext.libraryArtifactId  = "presenceviewer"
ext.libraryVersion = "0.0.1"
```

DittoExportLogs/build.gradle
```diff
@@ -3,6 +3,10 @@ plugins {
     id "org.jetbrains.kotlin.android"
 }
 
+ext.libraryGroupId = "live.ditto.dittoexportlogs"
+ext.libraryVersion = "0.0.1"
```
phatblat, line 7:
```suggestion
ext.libraryGroupId = "live.ditto"
ext.libraryArtifactId = "exportlogs"
ext.libraryVersion = "0.0.1"
```

DittoDiskUsage/build.gradle
```diff
@@ -3,10 +3,15 @@ plugins {
     alias libs.plugins.org.jetbrains.kotlin.android
 }
 
+
+ext.libraryGroupId = "live.ditto.dittodiskusage"
+ext.libraryVersion = "0.0.1"
```
phatblat, line 8: Access in Sonatype is based on the group ID. It would be easier to add new artifacts if we use our existing `live.ditto` as the group and then the library-specific name as the artifact ID (e.g. `diskusage`). This would result in full maven coordinates of `live.ditto:diskusage:<version>`. 

```suggestion
ext.libraryGroupId = "live.ditto"
ext.libraryArtifactId = "diskusage"
ext.libraryVersion = "0.0.1"
```

This depends on adding a `libraryArtifactId` extra var to `deploy.gradle`

Otherwise, I think we have to register a new organization and [verify that we own](https://central.sonatype.org/publish/requirements/coordinates/) the https://dittodiskusage.ditto.live domain.

phatblat (changes requested):
Group and artifact naming suggestions.

texasRanger09:
@phatblat Does this require that we push an update to the libraries in Maven as well now that I've added an artifactId?

anyercastillo:
I can see the following three options for package name:
- `live.ditto.dittopresenceviewer:DittoPresenceViewer:0.0.1` [used by this PR]
- `live.ditto:dittopresenceviewer:0.0.1`
- `live.ditto.tools:dittopresenceviewer:0.0.1`

No matter which one we choose, the @phatblat suggestion of define the `artifactId` is very important to prevent undesired packages name as a result of a module rename.

phatblat:
> @phatblat If we had already published a version of the tools without the ArtifactId, then do we just need to push an update to them? I.e this won’t create a new maven repo for each library now that it has an artifactId, right?

Without specifying the `artifactId` value, a name will be used automatically - typically the gradle project name. This is fragile because without `rootProject.name` set in `settings.gradle(.kts)`, the name will be the current directory name.

As for the versions previously published, they look broken to me. The 
https://s01.oss.sonatype.org/content/groups/public/live/ditto/ folder lists all maven "artifacts" published for the `live.ditto` group ID. Note that `ditto`, `ditto-cpp` and `dnssd` all contain a `maven-metadata.xml` file and folders for each version. The artifacts you've published contain only a subfolder, so I don't know if tools like gradle and maven will find these things that they need to find the actual artifacts.

texasRanger09:
@anyercastillo @phatblat I added the artifactID to each library and pushed an update of the libraries to Maven

phatblat:
The `0.0.2` versions look better. There is a `0.0.2` version folder under  https://s01.oss.sonatype.org/content/groups/public/live/ditto/dittodatabrowser/ and the [`maven-metadata.yml`](https://s01.oss.sonatype.org/content/groups/public/live/ditto/dittodatabrowser/maven-metadata.xml) file shows up in the right place.

texasRanger09:
I have updated all of the libraries and the readME to include the changes. 
If I could get another review cc/ @phatblat, @anyercastillo

README.md
```diff
@@ -53,7 +73,24 @@ DittoDataBrowser(ditto = ditto)
...
+dependencies {
+  implementation 'live.ditto:dittodatabrowser:0.0.2'
+}
+```
+
+Maven:
+```
+<dependency>
+    <groupId>live.ditto.</groupId>
+    <artifactId>DittoDataBrowser</artifactId>
```
phatblat, line 90:
```suggestion
    <artifactId>dittodatabrowser</artifactId>
```

README.md
```diff
@@ -103,9 +158,26 @@ Disk Usage allows you to see Ditto's file space usage.
...
+dependencies {
+  implementation 'live.ditto:dittodiskusage:0.0.2'
+}
+```
+
+Maven:
+```
+<dependency>
+    <groupId>live.ditto</groupId>
+    <artifactId>DittoDiskUsage</artifactId>
```
phatblat, line 176:
```suggestion
    <artifactId>dittodiskusage</artifactId>
```

app/build.gradle
```diff
@@ -26,10 +26,10 @@ dependencies {
...
     implementation platform(libs.androidx.compose.composeBom)
 
-    implementation project(path: ":DittoDiskUsage")
-    implementation project(path: ":DittoDataBrowser")
-    implementation project(path: ":DittoExportLogs")
-    implementation project(path: ':DittoPresenceViewer')
+    implementation libs.live.ditto.databrowser
+    implementation libs.live.ditto.exportlogs
+    implementation libs.live.ditto.presenceviewer
+    implementation libs.live.ditto.diskusage
```
phatblat, line 32: 😍 This looks so clean. I love version catalogs.

phatblat (approved):

Issue #22, opened 2024-01-20
PermissionsHealth


okdistribute:
- Show a green screen if 
  - permissions are granted for Ditto, 
  - Bluetooth is turned on, and 
  - LAN is turned on.
- Show a red screen if any of these are not true, and describe the problem on the screen.
  - Show a button which fixes the issue -- requests the permissions if necessary

PR #29, opened 2024-01-29
Introduce Health. Checks for Ditto Missing Permissions

anyercastillo:
Ticket: https://github.com/getditto/DittoAndroidTools/issues/22
This PR partially implements the ticket.
- permissions are granted for Ditto,

Other acceptances criteria will be implemented in future PRs.
- Bluetooth is turned on, and
- LAN is turned on.

### Video 
https://github.com/getditto/DittoAndroidTools/assets/5747647/710f5156-1097-4626-baf7-b57410192efa


DittoHealth/src/main/res/xml/provider_paths.xml
byoung-ditto, line 1: Maybe not necessary, perhaps a copy/paste leftover

app/src/main/java/live/ditto/dittotoolsapp/MainActivity.kt
```diff
@@ -37,7 +38,7 @@ class MainActivity : ComponentActivity() {
         lateinit var ditto: Ditto
         try {
             val androidDependencies = DefaultAndroidDittoDependencies(applicationContext)
-            val identity = DittoIdentity.OnlinePlayground(androidDependencies, appId = "YOUR_APPID", token = "YOUR_TOKEN", enableDittoCloudSync = true)
+            val identity = DittoIdentity.OnlinePlayground(androidDependencies, appId = "c5adf242-f3fc-4987-aac9-3cb79967274e", token = "812872a2-28a5-4ee7-b487-04f452eb6176", enableDittoCloudSync = true)
```
byoung-ditto, line 41: Is the AppId/token addition intentional?

DittoHealth/src/test/java/live/ditto/dittoexportlogs/ExampleUnitTest.kt
byoung-ditto, line 1: The  package path dittoexportlogs may be a copy/paste leftover

DittoHealth/src/main/java/live/ditto/health/usecase/GetDittoMissingPermissionsFlow.kt
```diff
@@ -0,0 +1,21 @@
...
+import live.ditto.transports.DittoSyncPermissions
+import kotlin.coroutines.coroutineContext
+
+class GetDittoMissingPermissionsFlow(
+    private val context: Context,
+) {
+    operator fun invoke(
+        timeoutMs: Long = 300
+    ) = flow {
+        while (coroutineContext.isActive) {
```
byoung-ditto, line 16: Looks good, wonder if in future we can do this without looping? Perhaps a permissions listener or Android Broadcast Receiver of some sort

DittoHealth/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+<!--    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>-->
+
+
+    <application>
+        <provider
```
byoung-ditto, line 8: Maybe not necessary, perhaps a copy/paste leftover

DittoHealth/src/androidTest/java/live/ditto/dittoexportlogs/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+@RunWith(AndroidJUnit4::class)
+class ExampleInstrumentedTest {
+    @Test
+    fun useAppContext() {
+        // Context of the app under test.
+        val appContext = InstrumentationRegistry.getInstrumentation().targetContext
+        assertEquals("live.ditto.dittoexportlogs.test", appContext.packageName)
```
byoung-ditto, line 22: Maybe wrong context/path? perhaps a copy/paste leftover

byoung-ditto:
When testing I was warned of 1 or 4 permissions in different cases

Install app, upon first launch click "Dont Allow" when prompted to allow permissions
	The app displays Red/Not Healthy with 4 missing permissions

In contrast install app, upon first launch click "Allow" when prompted to allow permissions
	The app displays Green/Healthy
	Now go into Android App settings, and manually remove the "Nearby devices" permission for "DittoToolsApp"
	Reopen the app, it displayed Red/Not Healthy with 1 mission permission

Havent investigated why it displays different list of permissions (1 vs 4) in these two cases, but could this being confusing for developers as well as app end users

Screenshot_20240129_194043
Screenshot_20240129_194314

byoung-ditto:
A couple final notes:

1) when "Nearby devices" is disallowed in the Android app settings, when opening the DittoToolsApp, a white screen is displayed for a few seconds. Recommend we verify this behavior is an artifact of the test app, and not of the "Ditto Health" module that will be embedded in customer apps

2) In some cases, I was prompted to allow Android Notifications. Even when tapping "Dont Allow" the Notifications permission was not listed in the Health Viewer, and did not seem to impact the Red vs Green status.

3) The UI/view is great for developers and tech support. If this is intended for end users, maybe consider adding a more human readable messages like those displaed in the Android App settings e.g. "Nearby devices" rather than "android.permission.NEARBY_WIFI_DEVICES" could be the main/large label, with the full permission string as a sublabel

DittoHealth/src/main/java/live/ditto/health/usecase/GetDittoMissingPermissionsFlow.kt
anyercastillo: Permissions are not observables in Android.
I am afraid is not other way to do it :(

DittoHealth/src/main/AndroidManifest.xml
anyercastillo: Yeah,
I copy/paste one of the modules and I did not pay attention to the manifest. 
My bad.

DittoHealth/src/androidTest/java/live/ditto/dittoexportlogs/ExampleInstrumentedTest.kt
anyercastillo: These tests have no values, I will delete the files.
thanks

app/src/main/java/live/ditto/dittotoolsapp/MainActivity.kt
anyercastillo: Hi, 
Good catch,
It is not intentional, I change those value to test the app. 
It should not be part of the PR. 
I will remove it.

Thanks

anyercastillo:
Thanks for the test!
Let me address some of your comments:

The module always prints the missing permissions returned by the Ditto SDK, it is hard for me to explain why the function returns 1 vs 4 permissions, I think is worth to take a look to the function implementation to understand the reason.

The white screen is the result of the MainScreen [instantiating Ditto on the Main thread](https://github.com/getditto/DittoAndroidTools/blob/99eadb2667e9f5782fc1f4c4d1e1e80ebe917cf7/app/src/main/java/live/ditto/dittotoolsapp/MainActivity.kt#L42). Please notice it is not possible for the component to render a white screen, once the composable is loaded by the navigation controller, it will render a green or red screen.
```
@Composable
private fun HealthScreen(state: HealthUiState) {
    if (state.isHealthy) {
        Healthy()
    } else {
        NotHealthy(state)
    }
}
```

DittoHealth/src/main/java/live/ditto/health/HealthScreen.kt
```diff
@@ -0,0 +1,143 @@
...
+            )
+
+        Spacer(modifier = Modifier.height(16.dp))
+
+        state.causes.forEach { cause ->
+            Cause(cause = cause)
+            Spacer(modifier = Modifier.height(16.dp))
+        }
+    }
+}
```
bplattenburg, line 91: Minor - I'm not sure how much effort its worth putting into a quick demo app like this, but does it make sense to usehealthy / unhealthy state from the `HealthUiState` here directly and collapse `Healthy()` and `NotHealthy(state:)` into a single composable that swaps the colors based on the health check? It seems like maybe there's some duplicated UI logic that could be simplified

DittoHealth/src/test/java/live/ditto/health/ExampleUnitTest.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Example local unit test, which will execute on the development machine (host).
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+class ExampleUnitTest {
+    @Test
+    fun addition_isCorrect() {
+        assertEquals(4, 2 + 2)
+    }
+}
```
bplattenburg, line 17: I think we can remove the sample test

bplattenburg (commented):
Looks great! I left a few minor comments but nothing blocking

DittoHealth/src/main/java/live/ditto/health/HealthScreen.kt
anyercastillo: I am not following you, 
will ping you in Slack, so we can talk more about it.

bplattenburg: After our slack conversation, there's not really any duplication going on here, and I need to get another coffee.

anyercastillo: @bplattenburg and I met on Slack and we agreed on the following.
- Move the logic to resolve texts from the composable to the UiState model
- Keep the existing structure of Healthy/NotHealthy composables

okdistribute:
Thanks Anyer!

Can we focus the language to be slightly more for an end-user? For example, 

* instead of "system is healthy" say "all permissions granted" / hard to know what system you're referring to
* When missing location permissions, is it possible to show a button that when pressed will force the operating system ask for permissions again?

DittoHealth/src/test/java/live/ditto/health/ExampleUnitTest.kt
anyercastillo: no op, for now.
until we figure out the path for other modules too.
if we decide to remove tests, then we should remove in all modules

anyercastillo:
Sure thing Rae

anyercastillo:
With the [last commit](https://github.com/getditto/DittoAndroidTools/pull/29/commits/40287b22988cd9230ae78dfb43f5d911dfacdd66) this is how looks the UI

https://github.com/getditto/DittoAndroidTools/assets/5747647/35e903f2-86ac-47d9-b8d4-f58bd8c79b9b

anyercastillo:
Let's wait for https://github.com/getditto/DittoAndroidTools/pull/30
Then I will rebase and fix any conflict

anyercastillo:
Note: The latest main contains the required code to publish artifacts. 
After rebasing on main, let's add a commit on top to address deployment.

okdistribute:
LGTM

marinakuksoditto (approved):
Looks great!

Issue #26, opened 2024-01-24
Android: HeartbeatTool


okdistribute:
Android version of the HeartbeatTool https://github.com/getditto/DittoSwiftTools/issues/73

PR #32, opened 2024-02-01
Heart beat tool

texasRanger09:
\* I will update this as it changes

Data model:
```
{
    _id: {
        *passed in by user + dittoPeerKey
    },
    interval: String,
    remotePeersCount: Int,
    lastUpdated: String (ISO-8601)
    sdk: String
    presence: {
        <peerKey>: {
            deviceName: String,
            isConnectedToDittoCloud: Bool,
            bluetooth: Int,
            p2pWifi: Int,
            lan: Int,
        },
        <peerKey>…,
        …
    }
}
```

Config file that user passes in:
```
// Provided with the Heartbeat tool
data class HeartbeatConfig(
    val id: Map<String, String>,
    val interval: Long,
    val collectionName: String,
    val metaData: Map<String, Any>
)

// User defines the values here
// Passed into Heartbeat tool
  val config = HeartbeatConfig(
      id:  Map<String, String>
      interval: Int,
      collectionName: String,
      metaData: Map<String, Any>
  )
```


DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -0,0 +1,149 @@
...
+
+fun getConnections(presence: Presence?): Map<String, Any> {
+
+    val connectionsMap: MutableMap<String, Any> = mutableMapOf()
+
+    presence?.connections?.forEach { connection ->
+        val connectionsTypeMap = getConnectionTypeCount(connection = connection)
+
+        val connectionMap: Map<String, Any?> = mapOf(
+            "deviceName" to connection.deviceName,
```
okdistribute, line 94: We talked about adding the peerKey

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -0,0 +1,149 @@
...
+
+    return presence
+}
+
+fun addToCollection(info: HeartbeatInfo, config: HeartbeatConfig, ditto: Ditto) {
+
+    val doc = mapOf(
+        "_id" to mapOf(
+            "deviceId" to config.metaData["deviceId"],
+            "locationId" to config.metaData["locationId"]
```
okdistribute, line 75: peerKey?

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -0,0 +1,149 @@
...
+    val id: Map<String, String>,
+    val lastUpdated: String,
+    val presence: Presence?,
+)
+
+data class Presence(
+    val totalConnections: Int,
+    val connections: List<DittoPeer>,
+)
+fun startHeartbeat(ditto: Ditto, config: HeartbeatConfig, onDataLog: (HeartbeatInfo) -> Unit): Handler {
```
anyercastillo, line 30: What do you think about returning a flow?
```
fun heartbeatFlow(
   ditto: Ditto,
   delayTimeMs Int = 30_000,
   ...
) = flow {
  while(isActive) {
      ...
      val info = HeartbeatInfo(...)
      emit(info)
      delay(delayTimeMs)
  }
}
```

Having a flow gives you the benefits of free cancellation once the flow is not collected.
Additionally because flows are cold, nothing is executed until it is observed.

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -0,0 +1,149 @@
...
+import live.ditto.DittoPeer
+import live.ditto.DittoPresenceObserver
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
+
+data class HeartbeatConfig(
+    val interval: Long,
+    val collectionName: String,
+    val metaData: Map<String, Any>
```
anyercastillo, line 17: What do you think about using a data class here instead of a map?

```
data class Metadata(
    val deviceId: String,
    val locationId: String,
    ...
)
```

With a data class in place we can make use of extension functions to convert it to a map.

```
fun Metadata.toMap() = mapOf(
    "deviceId": deviceId,
    "locationId": locationId,
    ...
)
```

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -0,0 +1,149 @@
...
+)
+
+data class HeartbeatInfo(
+    val id: Map<String, String>,
+    val lastUpdated: String,
+    val presence: Presence?,
+)
+
+data class Presence(
```
anyercastillo, line 26: Based on how this class is used, seems like `totalConnections` is a derived property from `connections`
Could you also consider rename "connections" to "peers"?

```
data class Presence(
   val peers: List<DittoPeer>,
) 
```

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
texasRanger09: It looks like the peerKey changes on each call to `presence.graph`. This means that adding the `peerKey` in the `_id` will cause a new document to be created each time.

anyercastillo: My two cents here.

PeerKey is an array of bytes.
Do you think makes sense to have a hash representation of the bytes?
A hash is at least more readable.

anyercastillo:
This work is so good, 
Thank you so much for taking the time and effort to get this done.

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
okdistribute: Yes it would be best to use the hash. The peerKey shouldn't change on each call that is strange

texasRanger09: Yep, realized my issue. I've converted it to a hash and it's working as expected

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
texasRanger09: The metadata was initially intended for passing in the id's (ex. locationId and deviceId) that would then create the composite id in the ditto collection. I've since updated the config to have an id key where the user will pass all of the id's they want to use to create their composite key. So it leaves me with not knowing what the metadata would be used for (in the current state of the heartbeat tool) so I might remove it til its needed.

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -0,0 +1,135 @@
...
+            "deviceName" to connection.deviceName,
+            "devicePeerKey" to byteArrayToHash(connection.peerKey),
+            "isConnectedToDittoCloud" to connection.isConnectedToDittoCloud,
+//            "totalPeers" to connection.connections.size,
+            "bluetooth" to connectionsTypeMap["bt"],
+            "p2pWifi" to connectionsTypeMap["p2pWifi"],
+            "lan" to connectionsTypeMap["lan"],
+            )
+
+        connectionsMap["peer ${connectionsMap.size + 1}"] = connectionMap
```
okdistribute, line 115: Please use the peerKey as the key here because otherwise it is not going to be consistent across updates

okdistribute (changes requested):

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    return presence
+}
+
+fun addToCollection(info: HeartbeatInfo, config: HeartbeatConfig, ditto: Ditto) {
+
+    val doc = mapOf(
+        "_id" to info.id,
+        "interval" to "${config.interval / 1000} sec",
+        "totalPeers" to (info.presence?.totalPeers ?: 0),
```
okdistribute, line 90: Can we call this `remotePeersCount`?

okdistribute (changes requested):

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -0,0 +1,135 @@
...
+        "lastUpdated" to info.lastUpdated,
+        "presence" to getConnections(info.presence)
+    )
+
+    ditto.store.collection(config.collectionName).upsert(doc)
```
okdistribute, line 95: Can you use DQL?

okdistribute (dismissed):

okdistribute:
Great work thanks Walker! Just a few comments and should be good to go

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
texasRanger09: Making the key for this map the peerKey, so no need to add an additional peerKey property.

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -0,0 +1,129 @@
...
+        )
+        addToCollection(info, config, ditto)
+        emit(info)
+    }
+}
+
+fun createCompositeId(configId: Map<String, String>, ditto: Ditto): Map<String, String> {
+    val compositeId: MutableMap<String, String> = configId.toMutableMap()
+    val presenceGraph = ditto.presence.graph
+    compositeId["dittoPeerKey"] = byteArrayToHash(presenceGraph.localPeer.peerKey)
```
bplattenburg, line 57: Assuming this is the same format, should we add the `pk` prefix here to align these to the keys used in Small Peer Info and elsewhere?

texasRanger09: I had it as `dittoPeerKey` since this will be added to their composite Id automatically. I wanted to give visibility that it was a ditto variable. Didn't know if users would be confused if it was just `pk` and why it was automatically being added to their composite Id.

okdistribute: Yeah let's try to align this with the `pk` prefix

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -0,0 +1,141 @@
...
+    return compositeId
+}
+
+fun byteArrayToHash(byteArray: ByteArray): String {
+    val md5 = MessageDigest.getInstance("md5")
+
+    return BigInteger(1, md5.digest(byteArray))
+        .toString(16)
+        .padStart(32, '0')
+}
```
bplattenburg, line 79: I think we still need to change the logic here to be a base64 encoding of the bytes (URL safe) with the string "pk" stuck on the front.

bplattenburg (changes requested):

bplattenburg:
Closes https://github.com/getditto/DittoAndroidTools/issues/26

bplattenburg (approved):

PR #33, opened 2024-02-03
Presence Degradation Reporter

anyercastillo:
Introduces Presence Degradation Reporter.
Documentation can be found in the README.md

## VIDEO
https://github.com/getditto/DittoAndroidTools/assets/5747647/b2191209-9fbd-4a80-b2fb-28d16c155c80

## ARTIFACTS
Sonatype artifacts are ready for deployment. [NOT DEPLOYED YET]

Screenshot 2024-02-12 at 11 22 33 AM


texasRanger09:
Would it be helpful to add a "Clear Remote Devices" button? It looks like the only way to clear the list of remote devices is to restart the app. I'm thinking this would mostly be helpful for clearing out devices that are no longer seen and you do not want them to clutter up the screen anymore.

texasRanger09:
We should display the `peerKey` value for each device as well. This should help with debugging since that value is used in the ditto logs.

anyercastillo:
> We should display the `peerKey` value for each device as well. This would help with debugging since that value is used in the ditto logs.

sounds like a very good idea.
thanks

anyercastillo:
peer-key

texasRanger09 (approved):

PR #42, opened 2024-03-19
update data model

texasRanger09:
```
 {
_id: , // peer key
secondsInterval: String, // how often does the heartbeat tool take a snapshot and store it here (in seconds)
lastUpdated: String (ISO-8601 UTC), // when this document was written. If this is farther in the past than the interval, this device is "offline"
sdk: String, // Ditto SDK version
presenceSnapshotDirectlyConnectedPeersCount: Int, // from presence graph snapshot
presenceSnapshotDirectlyConnectedPeers: { // from presence graph snapshot
: {
deviceName: String,
sdk: String,
isConnectedToDittoCloud: Bool,
bluetooth: Int,
p2pWifi: Int,
lan: Int,
},
…,
…
},
metaData: {}
}
```

### DIFF
--- a/DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
+++ b/DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
@@ -17,19 +17,20 @@ import java.util.Base64
 import java.util.concurrent.atomic.AtomicBoolean
 
 data class DittoHeartbeatConfig(
-    val id: Map<String, String>,
+    //remove:
+//    val id: Map<String, String>,
     val secondsInterval: Int,
     val collectionName: String,
     val metaData: Map<String, Any>? = null
 )
 
 data class DittoHeartbeatInfo(
-    val id: Map<String, String>,
+    val id: String,
     val lastUpdated: String,
     val metaData: Map<String, Any>?,
     val secondsInterval: Int,
-    val remotePeersCount: Int,
-    val peerConnections: Map<String, Any>,
+    val presenceSnapshotDirectlyConnectedPeersCount: Int,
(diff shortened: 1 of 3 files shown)


DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -17,19 +17,20 @@ import java.util.Base64
 import java.util.concurrent.atomic.AtomicBoolean
 
 data class DittoHeartbeatConfig(
-    val id: Map<String, String>,
+    //remove:
+//    val id: Map<String, String>,
```
bplattenburg, line 21: I think you meant to remove this instead of comment it out?

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -68,12 +69,13 @@ fun startHeartbeat(ditto: Ditto, config: DittoHeartbeatConfig): Flow<DittoHeartb
...
-    compositeId["pk"] = byteArrayToHash(presenceGraph.localPeer.peerKey)
-    return compositeId
-}
+//remove:
+//fun createCompositeId(configId: Map<String, String>, ditto: Ditto): Map<String, String> {
+//    val compositeId: MutableMap<String, String> = configId.toMutableMap()
+//    val presenceGraph = ditto.presence.graph
+//    compositeId["pk"] = byteArrayToHash(presenceGraph.localPeer.peerKey)
+//    return compositeId
+//}
```
bplattenburg, line 78: I think you meant to remove this instead of comment it out?

bplattenburg (changes requested):

bplattenburg (approved):

PR #46, opened 2024-03-27
Use default collection name

texasRanger09:
Sets `__dittotools_devices` as the default collection name


rajramsaroop (changes requested):
Update the `README.md` xD

We can remove references to setting the collection name since we removed it.

README.md
```diff
@@ -231,7 +231,6 @@ The Ditto Heartbeat tool allows you to monitor, locally or remotely, the peers i
 There are three values you need to provide to the Heartbeat:
```
rajramsaroop, line 231: Should be 4. I'd reword to say "these are the values we need to provide to the Heartbeat config" or something like that

rajramsaroop (changes requested):
Left comment regarding language on readme

rajramsaroop (approved):

PR #55, opened 2024-04-01
Use legacy builder to insert doc

texasRanger09:
Uses legacy builder to insert new docs. blocked til https://github.com/getditto/product/issues/159 is resolved.

### DIFF
--- a/DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
+++ b/DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
@@ -95,12 +95,10 @@ fun addToCollection(info: DittoHeartbeatInfo, config: DittoHeartbeatConfig, ditt
         "metaData" to metaData,
         "sdk" to info.sdk,
         "_schema" to info.schema,
-        "peerKey" to info.peerKey
+        "pk" to info.peerKey
     )
-    val query = "INSERT INTO $HEARTBEAT_COLLECTION_COLLECTION_NAME DOCUMENTS (:doc) ON ID CONFLICT DO UPDATE"
-    myCoroutineScope.launch {
-        ditto.store.execute(query, mapOf("doc" to doc))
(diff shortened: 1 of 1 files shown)


DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -95,12 +95,10 @@ fun addToCollection(info: DittoHeartbeatInfo, config: DittoHeartbeatConfig, ditt
         "metaData" to metaData,
         "sdk" to info.sdk,
         "_schema" to info.schema,
-        "peerKey" to info.peerKey
+        "pk" to info.peerKey
```
bplattenburg, line 98: Is this change intentional? I think we settled on spelled out `peerKey` for the name here

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -95,12 +95,10 @@ fun addToCollection(info: DittoHeartbeatInfo, config: DittoHeartbeatConfig, ditt
         "metaData" to metaData,
```
bplattenburg, line 95: I know this isn't changing as part of this PR, but I just noticed this is capitalized `metaData` here and `metadata` on the Swift version - we probably want to align those.

bplattenburg (changes requested):

DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
texasRanger09: Updated with this commit: https://github.com/getditto/DittoSwiftTools/pull/98/commits/8921870cc94bf6cc7901fedeacbd27d132cb99fb

rajramsaroop (approved):

PR #60, opened 2024-04-08
Add dependency constraint for ditto

texasRanger09:
Allow any ditto version greater than 4.4.0


bplattenburg:
> Changes look good.
> 
> Seems like we could have a case for making a Ditto Tools BOM? Could an issue/story be created to track this? CC: @bplattenburg

That makes sense to me, my only question would be whether thats a global change or per-tool given we publish the tools independently?

rajramsaroop:
We can still publish tools individually, but when we update the BOM, it would basically reference specific versions of tools. 

When using a BOM you can still pick and choose what tools you want to use, you just use the version compatible with the BOM.

Even when using a BOM, you can still choose to specify which version of the tool you want, so there's still flexibility if needed. The advantage of a BOM is that compatibility of the whole set is generally regarded to be good/tested.

texasRanger09:
I will merge this in and we can open another issue to cover the BOM.

cc/ @rajramsaroop @bplattenburg

bplattenburg:
> I will merge this in and we can open another issue to cover the BOM.
> 
> cc/ @rajramsaroop @bplattenburg

Sounds good! I broke the BOM changes out into https://github.com/getditto/DittoAndroidTools/issues/63, please let me know if we need more details there

PR #61, opened 2024-04-08
Bump dataBrowser version

texasRanger09:


DittoDataBrowser/src/main/java/live/ditto/dittodatabrowser/Documents.kt
```diff
@@ -97,12 +97,6 @@ fun Documents(collectionName: String, isStandAlone: Boolean) {
                     text = "No Docs",
                     textAlign = TextAlign.Start,
                     color = Color.Blue,
-                    modifier = Modifier
```
rajramsaroop, line 100: Why was this removed?

rajramsaroop (approved):
One question on the code, but everything related to changing the version looks good.

DittoDataBrowser/src/main/java/live/ditto/dittodatabrowser/Documents.kt
texasRanger09: It's on the "No Docs" text which means the user cannot click on the text to see the documents. There is nothing to modify when it shows "No Docs".

PR #70, opened 2024-04-26
Refactor of Presence Viewer

rajramsaroop:
Making some updates to the Presence Viewer.

Nothing should change about how the Presence Viewer is used, so existing consumers should be able to just pull the latest version without any code changes.


DittoPresenceViewer/src/main/java/ditto/live/dittopresenceviewer/DittoPresenceViewerWebView.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    init {
+        val assetLoader = WebViewAssetLoader.Builder()
+            .addPathHandler("/assets/dist/", WebViewAssetLoader.AssetsPathHandler(context))
+            .build()
+
+        layoutParams = LayoutParams(
+            ViewGroup.LayoutParams.MATCH_PARENT,
+            ViewGroup.LayoutParams.MATCH_PARENT,
+        )
```
bryansills, line 35: Can we make this a constructor parameter? I think having a default of `MATCH_PARENT, MATCH_PARENT` is fine, but it would be nice to give folks the configurability here.

bryansills (approved):
Minor suggestion. Otherwise, looks good.

bplattenburg:
@rajramsaroop Do we foresee any issues when we come back later and update this with the latest from the monorepo? (https://github.com/getditto/ditto/tree/main/android/dittopresenceviewer/src/main/java/live/ditto/dittopresenceviewer) I want to make sure we have a non-breaking path forward for consumers (that could include tweaking the Kotlin wrapper code currently in the monorepo, just want to make sure we have a clear path forward).

rajramsaroop:
With all that said, I think it would be good to test this PR's updated code, with the updated JavaScript code to ensure everything works as expected. I think I can do that in a separate branch/PR.

bplattenburg:
@rajramsaroop I think the full long term plan is still a bit TBD, but I expect there to only be a single source of truth (at least for the Kotlin & Swift wrappers), and that would be here. If you wanted to update the monorepo version to align to this, go for it! not sure all thats involved there but could be a great learning experience.

And yes, there are updates in the monorepo JS that we will bring over at a future date

texasRanger09:
@rajramsaroop When I run it I see that the devices aren't updating. It sets up the initial mesh when I first run it, but as devices come in and out of connection the UI is not being updated to reflect the changes.

texasRanger09:
Also, just as a side when you are testing it's important to remember to update the gradle file for the tools app to pull in the local library instead of the Maven one so that you can test the changes. I find myself forgetting to switch to the local library when testing, so just a reminder.

example:
```
//    implementation libs.live.ditto.presenceviewer
    implementation(project(":DittoPresenceViewer"))
```

rajramsaroop:
> @rajramsaroop When I run it I see that the devices aren't updating. It sets up the initial mesh when I first run it, but as devices come in and out of connection the UI is not being updated to reflect the changes.

PR updated. Seems like if I didn't call `close()` on the `DittoPresenceObserver` updating seemed to be slow or had the behavior you mention. I just tested on 2 physical devices and it worked correctly, updating immediately when there's any mesh changes.

rajramsaroop:
Actually it may be because I have a reference to the DittoPresenceObserver... Either way, calling close() makes sense as well and this fixes the issue.

texasRanger09 (approved):
Tested it and it's working. When a new transport comes online then the viewer is sometimes slow to update to show that new connection, but that does not seem related to this PR.

PR #77, opened 2024-05-17
fix gradle and build errors

rajramsaroop:
There were some build errors with the tools app related to gradle dependencies + error in code.

Should now be able to build and run the app.


rajramsaroop:
Question - does HealthMetrics have to be a separate module? Seems like it could just be a package in `DittoHeartbeat` module. They're related and metrics wouldn't be used anywhere else?

bplattenburg:
> Question - does HealthMetrics have to be a separate module? Seems like it could just be a package in `DittoHeartbeat` module. They're related and metrics wouldn't be used anywhere else?

on the Swift side its separate because the `HealthMetricProvider` protocol needs to be visible to other tools so that they can conform and provide health metrics. I _think_ the same thing makes sense here?

texasRanger09:
> Question - does HealthMetrics have to be a separate module? Seems like it could just be a package in DittoHeartbeat module. They're related and metrics wouldn't be used anywhere else?
> 
> on the Swift side its separate because the HealthMetricProvider protocol needs to be visible to other tools so that they can conform and provide health metrics. I think the same thing makes sense here?

Yeah, the idea is that HealthMetrics will be added to other tools. It will be used to add the given tools data to the Heartbeat tool. It is a seperate module, but we don't need to publish it to Maven as a seperate module. We just need to make sure it is included in the `.aar` for each module that uses it.

DittoHealthMetrics/build.gradle
```diff
@@ -11,15 +11,27 @@ apply from: "$rootProject.projectDir/gradle/deploy.gradle"
 apply from: "$rootProject.projectDir/gradle/android-common.gradle"
 
 android {
-    namespace 'live.ditto.dittohealthmetrics'
+    namespace "live.ditto.health"
```
texasRanger09, line 14: should we use `healthmetrics` for the namespace?

bplattenburg:
@zmarkan did https://github.com/getditto/DittoAndroidTools/pull/79 include these changes as well?

zmarkan:
@bplattenburg yes, it included everything apart from the last commit with namespace change - https://github.com/getditto/DittoAndroidTools/pull/77/commits/abf9ea971197a3742557dd70ade3e8c4774d8979

bplattenburg:
I think we're good to close this then? @rajramsaroop

DittoHealthMetrics/build.gradle
```diff
@@ -1,6 +1,6 @@
 plugins {
-    alias libs.plugins.com.android.library
-    alias libs.plugins.org.jetbrains.kotlin.android
+    id "com.android.library"
+    id "org.jetbrains.kotlin.android"
```
phatblat, line 3: What's the reason for moving away from version catalog entries for these? Were they causing an issue?

rajramsaroop: No issues, I just did a copy and paste without updating them to use version catalog.

rajramsaroop:
> I think we're good to close this then? @rajramsaroop

Gonna include the changes Ben mentioned, but otherwise Zan's PR covered everything else.

DittoHealthMetrics/build.gradle
```diff
@@ -11,7 +11,7 @@ apply from: "$rootProject.projectDir/gradle/deploy.gradle"
 apply from: "$rootProject.projectDir/gradle/android-common.gradle"
 
 android {
-    namespace "live.ditto.health"
+    namespace "live.ditto.dittohealthmetrics"
```
phatblat, line 14: I got the impression from @texasRanger09's [comment](https://github.com/getditto/DittoAndroidTools/pull/77#discussion_r1613727266) that this was changing to `live.ditto.healthmetrics`? Is that still desired?

```suggestion
    namespace "live.ditto.healthmetrics"
```

bplattenburg: Since this discussion originally happened, we've published this as `live.ditto.health`: https://central.sonatype.com/artifact/live.ditto/health.

Not sure if that changes what we want to do here?

bplattenburg: ...which seems to have resulted in https://github.com/getditto/DittoAndroidTools/issues/101 since we now have two `live.ditto.health` modules based on what @rajramsaroop found today

rajramsaroop:
I updated the namespace to match what I will update in my other PR to fix the duplicate class issue.

This PR also updates the build.gradle to use version catalog for the plugin imports.

phatblat (approved):

PR #78, opened 2024-05-21
Update tools to use peer key string

### LINKED ISSUE #35: Android - Update tools to use peerKeyString
Now that The 4.7.0 release of the Ditto SDK is live, we should change every tool that calculates the Ditto Peer to instead use the new `peerKeyString`.

Note this will require bumping the minimum Ditto SDK version for the tools to 4.7.0

rajramsaroop:
Closes #35 

Also included some other updates to fix gradle build issues.


DittoHealthMetrics/build.gradle
```diff
@@ -11,15 +11,27 @@ apply from: "$rootProject.projectDir/gradle/deploy.gradle"
...
+    implementation libs.androidx.compose.ui.ui
+    implementation libs.androidx.compose.ui.uiTooling
+    implementation libs.androidx.compose.ui.uiToolingPreview
+    implementation libs.androidx.navigation.navigationCompose
+    implementation libs.live.ditto.ditto
+    // Dependency constraint for ditto
+    constraints {
+        implementation(libs.live.ditto.ditto) {
+            version {
+                strictly "[4.5.0,)"
```
bplattenburg, line 30: Should this also bump to 4.7.0+?

bplattenburg:
We're going to hold off on merging this at this time to keep the tools support for Ditto SDK 4.5.x

DittoHealthMetrics/build.gradle
rajramsaroop: Yes good catch!

rajramsaroop: Question actually - do we want ALL tools to use a min. version of 4.7.0+? Or just the ones that use the peerKeyString?

bplattenburg: Hmm, good question - on the Swift side, it's all or nothing because the tools are all bundled as a single SPM package, so I didn't consider that.

If we can do these independently and really don't need anything from 4.7 for some tools, let's do it to make as many tools as possible as usable as possible.

rajramsaroop: Okay, I've updated the min version for the specific tools that use the peerkey string.

bplattenburg:
This can now be merged (pending merge conflict resolution)

bplattenburg (approved):

PR #81, opened 2024-05-28
add health metrics to permissions health

texasRanger09:


bplattenburg (approved):

DittoHealth/build.gradle
```diff
@@ -22,6 +22,7 @@ dependencies {
     implementation libs.androidx.compose.ui.uiTooling
     implementation libs.androidx.compose.ui.uiToolingPreview
     implementation libs.androidx.navigation.navigationCompose
+    implementation project(path: ':DittoHealthMetrics')
```
rajramsaroop, line 25: was this just for testing?

DittoHealth/src/main/java/live/ditto/health/HealthViewModel.kt
```diff
@@ -59,4 +61,25 @@ class HealthViewModel(
             it.copy(bluetoothEnabled = status)
         }
     }
+
+    override val metricName: String
```
rajramsaroop, line 65: Can we create a UseCase to create this health metric?

rajramsaroop (changes requested):

DittoHealth/build.gradle
texasRanger09: That was to include the Health metrics into the .aar for the tool.

rajramsaroop (approved):

PR #82, opened 2024-05-29
bump export logs version

texasRanger09:
Bumped version and published to Maven


bplattenburg (approved):

rajramsaroop (changes requested):
Need to update the README file as well

texasRanger09:
> Need to update the README file as well

The ReadMe was not tracking latest versions. It shows the dependency information with an example version, but we have not been using that as the latest version reference. Do we want to update the ReadMe to include the latest versions for each tool?

zmarkan:
I used to write ReadMe files with dependency versions extracted like this:

```
  implementation 'live.ditto:dittopresenceviewer:YOUR_LIBRARY_VERSION'
``` 

... And then specify the latest lib version at the top somewhere and link to GH tags/releases (if we tie to that) and/or maven central. 

Just a thought - might make maintaining a little easier?

zmarkan (approved):

rajramsaroop:
> I used to write ReadMe files with dependency versions extracted like this:
> 
> ```
>   implementation 'live.ditto:dittopresenceviewer:YOUR_LIBRARY_VERSION'
> ```
> 
> ... And then specify the latest lib version at the top somewhere and link to GH tags/releases (if we tie to that) and/or maven central.
> 
> Just a thought - might make maintaining a little easier?

I like this idea since our tools are currently different versions and it would get annoying to update every time. I know most developers will know what's actually latest (if not Android studio will tell them lol) so this makes sense.

rajramsaroop (approved):

PR #83, opened 2024-05-30
We/diskusage healthmetrics

### LINKED ISSUE #75: Android - Disk Usage HealthMetricProvider
Add onto the Disk Usage tool the ability for it to be a HealthMetricProvider, with the following information:

`isHealthy` is `true` if the sum of `ditto_store` and `ditto_replication` is <= 2 GB (2147483648 bytes), `false` otherwise
`details` includes the full size list as shown in the UI

texasRanger09:
Closes #75

### DIFF
--- a/DittoDiskUsage/src/main/java/com/example/dittodiskusage/usecase/GetDiskUsageMetrics.kt
+++ b/DittoDiskUsage/src/main/java/com/example/dittodiskusage/usecase/GetDiskUsageMetrics.kt
@@ -9,8 +9,9 @@ import com.example.dittodiskusage.TOTAL_SIZE
 import com.example.dittodiskusage.TWO_GIGABYTES_IN_MEGABYTES
 import live.ditto.dittohealthmetrics.HealthMetric
 
-class GetDiskUsageMetrics {
+class GetDiskUsageMetrics() {
     val metricName: String = METRIC_NAME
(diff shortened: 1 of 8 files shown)


DittoDiskUsage/src/main/java/com/example/dittodiskusage/usecase/GetDiskUsageMetrics.kt
```diff
@@ -0,0 +1,52 @@
...
+        return HealthMetric(
+            isHealthy = isHealthy,
+            details = details
+        )
+    }
+
+    private fun healthCheckSize(dittoStoreSize: Int, dittoReplicationSize: Int): Boolean {
+        val totalMB = bytesToMegabytes(dittoStoreSize + dittoReplicationSize)
+
+        if(totalMB <= TWO_GIGABYTES_IN_MEGABYTES) {
```
bplattenburg, line 39: Can we make the "unhealthy size" configurable but defaulting to 2GB?

bplattenburg (changes requested):

DittoDiskUsage/src/main/java/com/example/dittodiskusage/usecase/GetDiskUsageMetrics.kt
texasRanger09: Where should they be able to set that? Through the Heartbeat tool?

bplattenburg (approved):

Issue #80, opened 2024-05-24
Setup CI to publish new versions automatically


bplattenburg:

zmarkan:
Creds & package signing key are in 1password. 
Sync with @texasRanger09 and @rajramsaroop Wednesday 29 May

zmarkan:
I've been having lots of trouble getting Gradle to sign and upload archives to Maven Central. 
Tried a few different signing plugins and approaches  with limited success. 
Syncing with Raj later today to see if he can spot something obvious I may have missed

bplattenburg:
This can be closed, the work is done and being used today, but we missed closing the issue here.

PR #86, opened 2024-06-04
Create workflow actions for automated releasing

zmarkan:
New release and snapshots actions for automated publishing to Maven Central. Fulfills https://github.com/getditto/DittoAndroidTools/issues/80 

Release workflow triggers on new release publish in GitHub - all projects with the same version deployed at the same time - so we can keep consistency.  Signing is automated, using GitHub's secrets.

Snapshots workflow currently doesn't work but would ideally trigger nightly. Leaving the script in there to eventually debug & fix. 

Let me know if it makes sense - happy to walk you folks through it & answer any Qs. 

And most importantly thanks Raj for your help looking at error logs!!!


rajramsaroop:
Sorry if this is a silly question as I'm not that familiar with GHA. Can anyone with a GH account run the GHA, or do we need to setup permissions so that only certain people/level of access can publish?

rajramsaroop:
I think we should add either a `RELEASING.md` or something in Notion to describe how to release to describe the workflow for updating the versions then publishing etc.

rajramsaroop (commented):
Scripts look good to me, just left 2 comments.

zmarkan:
> Sorry if this is a silly question as I'm not that familiar with GHA. Can anyone with a GH account run the GHA, or do we need to setup permissions so that only certain people/level of access can publish?

At the moment, anyone with write access to the repo can publish a release which will trigger the release action. Seemed like a sensible default option.
If we need to, we can change permissions / triggers so that only members of certain groups can do it. Happy to hear others' thoughts?

> I think we should add either a `RELEASING.md` or something in Notion to describe how to release to describe the workflow for updating the versions then publishing etc.

Great idea - I'll write it up this morning. Can draft it in notion then either move to the repo or keep there.

zmarkan:
I documented everything - the process and the scripts. 
I think this should cover everything: https://www.notion.so/getditto/Publishing-Automation-f3ff687ee9f24e059bbc85af716f2d3d

bplattenburg:
@zmarkan Thanks for documenting! I think that is the right place to leave those docs given this is a public repo

rajramsaroop (approved):
Great work!

I think the current permissions are fine, but we can revisit later if anyone wants something different.

Notion doc is a good place too.

okdistribute:
Nice!

bplattenburg (approved):

zmarkan:
So do we just merge now it's been approved? Or is there a step in the flow that I'm missing?

rajramsaroop:
> So do we just merge now it's been approved? Or is there a step in the flow that I'm missing?

Yes you can merge!

PR #88, opened 2024-06-07
Default DiskUsage HealthMetric unhealthy limit to 500MB instead of 2GB

### LINKED ISSUE #87: Update Disk Usage HealthMetric unhealthy value
Update the default "unhealthy" size from 2GB to 500MB based on recent observations

bplattenburg:
Also includes a minor refactor to align the logic flow and bytes math to match how DittoSwiftTools does things (everything in bytes)

Closes #87


app/src/main/java/live/ditto/dittotoolsapp/HeartbeatView.kt
```diff
@@ -28,7 +28,6 @@ fun ShowHeartbeatData(ditto: Ditto) {
     var heartbeatInfo by remember { mutableStateOf<DittoHeartbeatInfo?>(null) }
     var healthMetricProviders: MutableList<HealthMetricProvider> = mutableListOf()
     val diskUsageViewModel = DiskUsageViewModel()
-    diskUsageViewModel.isHealthyMBSizeLimit = 2048 //2GB worth of data
```
bplattenburg, line 31: Removed because this is already the default value

DittoDiskUsage/src/main/java/com/example/dittodiskusage/DiskUsageViewModel.kt
```diff
@@ -28,11 +27,11 @@ class DiskUsageViewModel(
     /* Private mutable state */
     private val _uiState = MutableStateFlow(DiskUsageState())
 
-    /// The size over which disk usage is considered unhealthy when used as a `HealthMetric` with the heartbeat tool (this only considers `ditto_store` and `ditto_replication`). Defaults to 2GB
-    var isHealthyMBSizeLimit: Int = TWO_GIGABYTES_IN_MEGABYTES
+    /// The size over which disk usage is considered unhealthy when used as a `HealthMetric` with the heartbeat tool (this only considers `ditto_store` and `ditto_replication`). Defaults to 500MB
+    var unhealthySizeInBytes: Int = FIVE_HUNDRED_MEGABYTES_IN_BYTES
```
rajramsaroop, line 31: this variable is unused

DittoDiskUsage/src/main/java/com/example/dittodiskusage/usecase/GetDiskUsageMetrics.kt
```diff
@@ -6,12 +6,12 @@ import com.example.dittodiskusage.DiskUsageState
...
 import com.example.dittodiskusage.ROOT_PATH
 import com.example.dittodiskusage.TOTAL_SIZE
-import com.example.dittodiskusage.TWO_GIGABYTES_IN_MEGABYTES
+import com.example.dittodiskusage.FIVE_HUNDRED_MEGABYTES_IN_BYTES
 import live.ditto.dittohealthmetrics.HealthMetric
 
 class GetDiskUsageMetrics() {
     val metricName: String = METRIC_NAME
-    var isHealthyMBSizeLimit: Int = TWO_GIGABYTES_IN_MEGABYTES
+    var unhealthySizeInBytes: Int = FIVE_HUNDRED_MEGABYTES_IN_BYTES
```
rajramsaroop, line 14: you would want to pass this in as a parameter of `execute()`.

rajramsaroop (changes requested):

bplattenburg:
@rajramsaroop I tried to refactor a bit to address your feedback and bring this closer to how the `DittoSwiftTools` version works

DittoDiskUsage/src/main/java/com/example/dittodiskusage/DiskUsageViewModel.kt
bplattenburg: This is now used

DittoDiskUsage/src/main/java/com/example/dittodiskusage/usecase/GetDiskUsageMetrics.kt
bplattenburg: This is now part of `DiskUsageState` which is already passed there

rajramsaroop: How would a client set their own unhealthy size?

bplattenburg: It gets passed in from whatever they set in `DiskUsageViewModel.kt` every time the callback is triggered (which I just realized was private, fixed now)

rajramsaroop:
> @rajramsaroop I tried to refactor a bit to address your feedback and bring this closer to how the `DittoSwiftTools` version works

In general (personally) I think the same API's but for different platforms should match (in terms of naming, input/outputs etc), but anything that might be platform specific should look like what a developer of that platform expects. I think what we have here is OK for now (since we're favoring speed over perfection lol).

rajramsaroop (approved):

bplattenburg:
@rajramsaroop when you get a chance, can you rough out the refactor you think makes sense into its own issue for this repo? Its worth revisiting and making sure we've decoupled the UI and HealthMetricProvider portions of this tool fully.

Issue #91, opened 2024-06-10
Clean up unused version values in module level build.gradle files


zmarkan:
Gradle files have a number of unused `version` values set, that are overridden by the release CI pipeline.
They need to be removed to avoid confusion

Issue #90, opened 2024-06-10
Add the ability to publish to Maven locally


rajramsaroop:
Being able to publish locally will allow developers to import their local changes into other Android projects locally that are outside the tools repo. This is useful as it allows developers to properly test their changes.

Right now if you try to run `./gradlew publishToMavenLocal` you run into signing issues. Let's fix that and any other issues that might arise so that we can publish locally.

zmarkan:
That's a good shout.
The place to look would be the `signing` block in `gradle/deploy.gradle` - needs to omit anything apart from the maven release build triggers.

PR #92, opened 2024-06-13
Gradle improvements to local publishing and cleanup

zmarkan:
- Allow running `publishToMavenLocal` without signing credentials being set. Fixes https://github.com/getditto/DittoAndroidTools/issues/90. CI workflow now sets a special env var `ORG_GRADLE_PROJECT_PUBLISH_TO_MAVEN_CENTRAL` that is checked by the signing plug in in.
- Clean up gradle scripts - remove redundant `version` and `group` vars that are set in `gradle/deploy.gradle`. Fixes https://github.com/getditto/DittoAndroidTools/issues/91


bplattenburg (approved):
Looks good to me but I'm not an Android/Gradle expert

Issue #104, opened 2024-06-24
Update dependencies for demo app


rajramsaroop:
Today the demo app points to the local projects as some tools were not published to maven. 

We should update the dependencies in `build.gradle` of the demo app to point to the maven dependencies so it can be used as an up to date example of how to implement tools.

okdistribute:
The original thinking for pointing the dependencies to be local is for development. It helped us iterate faster on testing changes in the tools

rajramsaroop:
Currently there is a mix of some tools being pulled in from maven, others local. This would make it more consistent. There were also todo comments left for updating the imports to use maven as well.

If someone wanted to iterate and quickly test there's nothing to stop someone from commenting out the maven import and importing the local project. This works well for testing something in the demo app.

Zan also completed some work previously that allowed publishing to maven locally so that we can also test the tools changes locally in other projects, not just the tools project.

So even after these changes it's still very easy to use the local versions for testing and I like the idea of including something in the readme on how to do this. The steps themselves for both these options are simple and should take minutes to get up and running.

PR #107, opened 2024-06-25
Cleanup demo app dependencies. Add local testing instructions

rajramsaroop:
Fixes https://github.com/getditto/DittoAndroidTools/issues/104

Cleaned up the dependencies.

Also added instructions if someone wanted to test tools changes either locally or in an external project.

Added the ability to set the local version number in `local.properties` file (since the file doesn't get committed, less chance of something changing that we don't want). When running `./gradlew publishToMavenLocal` it will use the version set there. This should NOT affect publishing to Maven or the Github action - those should work as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,10 +409,9 @@ remove/comment out: `implementation libs.live.ditto.databrowser`
 
 ### Testing in an External Project
 
-1. In your `local.properties` file for `DittoAndroidTools` add a `LIBRARY_VERSION` variable with a version that doesn't exist (e.g. 82.0.0). Do NOT use quotes. The entry should look something like: `LIBRARY_VERSION=82.0.0`
-2. Run `./gradlew publishToMavenLocal`
-3. In your external project add the `mavenLocal()` entry to your list of repository sources
-4. When importing a tool, use the version you specified earlier
+1. Run `./gradlew publishToMavenLocal`
+2. In your external project add the `mavenLocal()` entry to your list of repository sources
+3. When importing a tool, the version will be `SNAPSHOT`
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -52,6 +52,23 @@ ditto.onlinePlayground.appId="YOUR_APPID"
 ditto.onlinePlayground.token="YOUR_TOKEN"
 ```
 
+## Testing Changes Locally
+
+There are two ways you can test things locally. Either in the demo app, or in an external project.
+
+### Testing in the Demo App Locally
+
+To test your changes to a module in the demo app, simply import the local module, making sure to comment out/delete the version catalog import. For example, if you made changes to DittoDataBrowser you would
+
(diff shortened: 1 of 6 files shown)


README.md
rajramsaroop, line 1: Wasn't sure if the testing section should go before or after the list of tools. The primary reason for coming to this repo will most likely be to integrate with the tools, so I was thinking the testing section could go after?

rajramsaroop: Also please let me know if the testing steps make sense. They should be pretty trivial and something that most android/kotlin/java devs are used to, but let me know!

bplattenburg: Having these at the end after the list of tools makes sense to me, and as somebody with limited Android experience I think this is really clear already.

bplattenburg (approved):
LGTM

bplattenburg (approved):

build.gradle
```diff
@@ -5,3 +5,22 @@ plugins {
...
+    def reader = propFile.newReader()
+    try {
+        prop.load(reader)
+    } finally {
+        reader.close()
+    }
+    return prop.get(key)
+}
+
+ext.getProperty = { name, defValue ->
```
zmarkan, line 22: I know this was just moved around from a different gradle script, but I'm not convinced that having two similarly worded methods `findProperty` and `getProperty` are the right way to go. 
I think using a more specific name like `getValueFromLocalProperties` might be better.

zmarkan (approved):
It's OK - just left a comment regarding `getProperty` that retrieves from `local.properties`.

build.gradle
rajramsaroop: I like that idea too, I'll update the PR.

gradle/deploy.gradle
```diff
@@ -37,7 +37,9 @@ afterEvaluate {
                 from components.release
                 artifactId libraryArtifactId
                 groupId = 'live.ditto'
-                version = findProperty("LIBRARY_VERSION")
+
+                def versionPropertyName = "LIBRARY_VERSION"
```
rajramsaroop, line 41: I'm wondering if no LIBRARY_VERSION can be found I just set to "0.0.0" as a default?

rajramsaroop: It means one less step in publishing locally.

zmarkan: That's a good shout - maybe just use a different string to avoid potential confusion with real semver? `SNAPSHOT` or something like that. 

If you do it also keep in mind the order of precedence for properties - https://docs.gradle.org/current/userguide/build_environment.html#priority_for_configurations

Currently the pipeline injects the prop as an environment variable which has the lowest precedence according to their docs - so if you stick the hardcoded test version in `gradle.properties` it will always take it from there.

zmarkan: I think that should work!

rajramsaroop: Cool, I'll update the PR and readme

PR #109, opened 2024-07-11
Update diskUsage image

### LINKED ISSUE #108: DiskUsage - make ditto directory exporting more obvious
Today, the Disk Usage tool allows users to export the Ditto Directory but it is not well documented and the "folder" icon at the top right is not obviously an action the user can take.

We should:

1. Document the export functionality clearly in the README
2. Replace the icon with a labelled button called something like "Export"

texasRanger09:
fixes #108 
See also #110


marinakuksoditto (approved):
Looks good!

bplattenburg (approved):

PR #110, opened 2024-07-11
Update Export Data icon and ReadMe

### LINKED ISSUE #108: DiskUsage - make ditto directory exporting more obvious

texasRanger09:
fixes #108 
See also #109 

Updates the ReadMe to include more data.
New Icon for Exporting data
diskUsage


marinakuksoditto (approved):
Looks good! Except line 55: "There are five components in this package: ", but it looks like there are actually 7 components.

bplattenburg (approved):

PR #111, opened 2024-08-14
Tools Entry Point

### LINKED ISSUE #24: Android: Provide a single entrypoint view that allows customers to get new tools without adding each tool individually
* Menu list of the tools by name.
* Click the tool and see the full-view
* Easily switch between tools with a visible menu

rajramsaroop:
Fixes #24 

Provides a Composable interface that clients can implement within their app to quickly get all of the tools. Only need to import one dependency `live.ditto:dittotoolsviewer:TOOLS_VERSION` and then include the below Composable providing a Ditto instance.

```
DittoToolsViewer(
     ditto = <your ditto instance>
)
```

Optional parameters include `modifier` if the layout needs to be adjust on client integrations.
`onExitTools` lambda function that will be executed when the "Exit Tools" button is tapped. Can use this to do any back navigation or dismiss the tools UI on the client integration.

All other configuration steps for tools still apply. For example if you want Data Exporter to work, you still need to configure a log location (as described in the current readme).

UI was reused from tools demo app.

### Sample integrations
- You can run the demo app included in the project to see what the interface looks like
- I have a [PR](https://github.com/getditto/demoapp-chat/pull/128) for the chat demo app that is integrating it locally. Be sure to publish tools with `./gradlew publishToMavenLocal`. 

Video demo of it running in the chat app:

[chat_tools_entry_demo.webm](https://github.com/user-attachments/assets/029bedd4-3829-4170-b65f-71be9361ea25)


zmarkan:
Picking this up now

bplattenburg (approved):

zmarkan:
"new tools who dis" 👏 🤣

zmarkan (approved):

PR #120, opened 2024-09-03
Update grade dependencies and fix for some crashes

rajramsaroop:
Now that we've published all tools we should be pulling in the dependencies using gradle as opposed to specifying the local project(s).

I also noticed a few crashes when integrating with apps outside the project. Export logs and Heartbeat for example. In it's current state you can reproduce those crashes by pulling in Tools Viewer 2.1.0 (latest published) and trying to use Export Logs or Heartbeat. I tested this with the demo app, chat app and Android PoS demo [of which I am working on locally].

To see it fixed, run `./gradlew publishToMavenLocal` for tools, then pull in the local tools version (`SNAPSHOT`).

Export Logs crashed due to a null value being returned from a predicate.
Heartbeat crashed because gradle couldn't figure out which class to use for `DittoCollection.upsert()`.

Once these fixes are in, I'd like to publish a 2.1.1. version.

### DIFF
--- a/DittoDiskUsage/build.gradle
+++ b/DittoDiskUsage/build.gradle
@@ -21,7 +21,7 @@ dependencies {
     implementation libs.androidx.compose.ui.ui
     implementation libs.androidx.compose.ui.uiToolingPreview
     implementation libs.androidx.navigation.navigationCompose
-    implementation project(path: ':DittoHealthMetrics')
+    implementation libs.live.ditto.healthmetrics
     implementation libs.live.ditto.ditto
     // Dependency constraint for ditto
     constraints {
@@ -31,7 +31,7 @@ dependencies {
             }
         }
     }
-    implementation project(":DittoExporter")
+    implementation libs.live.ditto.exporter
 
     implementation platform(libs.androidx.compose.composeBom)
 
--- a/DittoDiskUsage/src/main/java/live/ditto/dittodiskusage/usecase/GetDiskUsageMetrics.kt
+++ b/DittoDiskUsage/src/main/java/live/ditto/dittodiskusage/usecase/GetDiskUsageMetrics.kt
@@ -16,7 +16,7 @@ class GetDiskUsageMetrics {
     fun execute(currentState: DiskUsageState): HealthMetric {
 
         val dittoStoreSize: Int =
-            currentState.children.first { shortRelativePath(it.relativePath) == DITTO_STORE }.sizeInBytes
+            currentState.children.firstOrNull { shortRelativePath(it.relativePath) == DITTO_STORE }?.sizeInBytes ?: 0
         val dittoReplicationSize: Int = currentState.children.firstOrNull {
             shortRelativePath(it.relativePath) == DITTO_REPLICATION }?.sizeInBytes ?: 0
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -9,8 +9,8 @@ androidx-navigation = "2.5.3"
 androidx-test-ext = "1.1.5"
 androidx-webkit = "1.7.0"
 datastorePreferences = "1.0.0"
-ditto = "4.7.4"
-live-ditto-tools = "2.0.0"
+ditto = "4.8.0"
+live-ditto-tools = "2.1.0"
 
 junit = "4.13.2"
 kotlin-gradle-plugin = "1.8.10"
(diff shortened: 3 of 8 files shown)


DittoDiskUsage/src/main/java/live/ditto/dittodiskusage/usecase/GetDiskUsageMetrics.kt
```diff
@@ -16,7 +16,7 @@ class GetDiskUsageMetrics {
     fun execute(currentState: DiskUsageState): HealthMetric {
 
         val dittoStoreSize: Int =
-            currentState.children.first { shortRelativePath(it.relativePath) == DITTO_STORE }.sizeInBytes
+            currentState.children.firstOrNull { shortRelativePath(it.relativePath) == DITTO_STORE }?.sizeInBytes ?: 0
```
zmarkan, line 19: Huh, I thought this was fixed already, last time I was fighting with dependencies to review something

DittoDiskUsage/build.gradle
```diff
@@ -21,7 +21,7 @@ dependencies {
     implementation libs.androidx.compose.ui.ui
     implementation libs.androidx.compose.ui.uiToolingPreview
     implementation libs.androidx.navigation.navigationCompose
-    implementation project(path: ':DittoHealthMetrics')
+    implementation libs.live.ditto.healthmetrics
```
zmarkan, line 24: Not something we need to change but I'm now wondering whether it's possible to programmatically define where the deps are coming from - so the CI build would always default to the published gradle library dependencies vs project path dependencies.

settings.gradle
```diff
@@ -8,6 +8,7 @@ pluginManagement {
 dependencyResolutionManagement {
     repositoriesMode.set(RepositoriesMode.FAIL_ON_PROJECT_REPOS)
     repositories {
+        mavenLocal()
```
zmarkan, line 11: I think this one makes sense to stay here

zmarkan (dismissed):
Also make sure to update the README - now the `mavenLocal` dependency step is not required any more!

rajramsaroop:
> Also make sure to update the README - now the `mavenLocal` dependency step is not required any more!

It's still required for external projects

DittoDiskUsage/build.gradle
rajramsaroop: I think now that we're pointing to the gradle version instead of the local project, this should be good enough for now.

DittoDiskUsage/src/main/java/live/ditto/dittodiskusage/usecase/GetDiskUsageMetrics.kt
rajramsaroop: This was basically saying it didn't find anything matching the predicate for `DITTO_STORE `, which means that the directory doesn't exist or maybe we're looking in the wrong place?

Doing a null check means that if it fails at least it won't crash.

bplattenburg: It might be worth revisiting this after SDK 4.9.0, there are some stability improvements around the disk usage tooling in the SDK.

This is a great answer for now then.

DittoDiskUsage/build.gradle
bplattenburg: @zmarkan I like that thought longer term, maybe we log an issue to investigate that further in the future?

gradle/libs.versions.toml
```diff
@@ -9,8 +9,8 @@ androidx-navigation = "2.5.3"
 androidx-test-ext = "1.1.5"
 androidx-webkit = "1.7.0"
 datastorePreferences = "1.0.0"
-ditto = "4.7.4"
-live-ditto-tools = "2.0.0"
+ditto = "4.8.0"
```
bplattenburg, line 12: Can we make this SDK dependency upgrade its own PR? That way it shows up easily in the generated changelog and is easier to revert if something breaks.

Also, we need to remember this is a breaking change in terms of SemVer and requires a major version bump when we next release

bplattenburg (changes requested):

gradle/libs.versions.toml
rajramsaroop: PR updated

bplattenburg (approved):

PR #137, opened 2024-11-13
Update logExporter to use rotating logs

### LINKED ISSUE #59: Android - Support Ditto automatic rotating logs in the Log Exporter tool
Today, the Log Exporter tool requires that users turn on Ditto logs to a local file, which they then have to manage to it does not grow unbounded.

Since this tool was initially created, we created an automatic log rotation system which is on by default - the log exporter should ideally use this system instead and not require users to manage an additional log file in app code

zmarkan:
Fixes https://github.com/getditto/DittoAndroidTools/issues/59 

now ready for review, and hopefully publishing


DittoExportLogs/build.gradle
```diff
@@ -18,6 +18,7 @@ dependencies {
     implementation libs.androidx.compose.material3.material3
     implementation libs.androidx.compose.ui.ui
     implementation libs.androidx.navigation.navigationCompose
+    implementation(libs.live.ditto.ditto)
```
rajramsaroop, line 21: You'll want to merge the latest main as it resolves some of these dependency issues, specifically with the ditto sdk

DittoExportLogs/src/main/java/live/ditto/dittoexportlogs/ExportLogs.kt
```diff
@@ -1,26 +1,116 @@
...
+    val filePath = File(downloadsDir, fileName).absolutePath
+
+    AlertDialog(
+        modifier = modifier,
+        onDismissRequest = onDismiss,
+        title = {
+            Text(title)
+        },
+        text = {
+            Text("Logs will be exported to: \nDownloads/$fileName\"")
```
rajramsaroop, line 61: Let's stick this bad boy in a strings.xml

DittoExportLogs/src/main/java/live/ditto/dittoexportlogs/ExportLogs.kt
```diff
@@ -1,26 +1,116 @@
...
+                            Toast.makeText(context, toastText, Toast.LENGTH_LONG).show()
+                        }
+                        catch (e: DittoError.IoError) {
+                            Log.e("DittoTools LogExporter", e.reason.toString())
+                            Toast.makeText(context, "Error exporting logs: ${e.message}", Toast.LENGTH_LONG).show()
+                        }
+                        finally {
+                            isLoading = false
+                            onDismiss()
+                        }
```
rajramsaroop, line 85: Not sure what sort of time crunch we are on here, but I'd prefer this to be in a view model

DittoToolsViewer/build.gradle.kts
```diff
@@ -34,7 +34,8 @@ dependencies {
 
     implementation(libs.live.ditto.ditto)
     implementation(libs.live.ditto.databrowser)
-    implementation(libs.live.ditto.exportlogs)
+//    implementation(libs.live.ditto.exportlogs)
+    implementation(project(":DittoExportLogs"))
```
rajramsaroop, line 38: 1. Don't forget to remove this before merging
2. This creates another issue that I think I've added to the tools backlog (to be figured out another time?), but if we uncomment out the code, it won't be pointing to the latest version of the tool - https://github.com/getditto/DittoAndroidTools/issues/123

DittoExportLogs/src/main/java/live/ditto/dittoexportlogs/ExportLogs.kt
```diff
@@ -1,26 +1,116 @@
...
+                onClick = {
+
+
+                    coroutineScope.launch {
+                        isLoading = true
+                        try {
+                            val sizeWritten = DittoLogger.exportToFile(filePath)
+
+                            val sizeWrittenInKB = sizeWritten / 1024u
+                            val toastText = "Exported file size: $sizeWrittenInKB kB"
```
rajramsaroop, line 75: Put this string into the strings.xml

DittoExportLogs/src/main/java/live/ditto/dittoexportlogs/ExportLogs.kt
```diff
@@ -1,26 +1,116 @@
...
+                            Toast.makeText(context, toastText, Toast.LENGTH_LONG).show()
+                        }
+                        catch (e: DittoError.IoError) {
+                            Log.e("DittoTools LogExporter", e.reason.toString())
+                            Toast.makeText(context, "Error exporting logs: ${e.message}", Toast.LENGTH_LONG).show()
```
rajramsaroop, line 80: move this to a strings.xml

DittoExportLogs/src/main/java/live/ditto/dittoexportlogs/ExportLogs.kt
```diff
@@ -1,26 +1,97 @@
...
+
+    val now = LocalDateTime.now().format(DateTimeFormatter.ofPattern("yyyyMMdd_HHmmss"))
+    val appName = context.applicationInfo.loadLabel(context.packageManager).toString()
+
+    val fileName = "$appName-dittologs-$now.jsonl.gz"
+    val downloadsDir = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS);
+    val filePath = File(downloadsDir, fileName).absolutePath
+
+    AlertDialog(
+        modifier = modifier,
```
rajramsaroop, line 51: We can delete this parameter here as we never change it. And then remove the `val modifier = Modifier` on line 39 above.

DittoExportLogs/src/main/java/live/ditto/dittoexportlogs/ExportLogs.kt
```diff
@@ -1,26 +1,97 @@
...
+        text = {
+            Text(stringResource(R.string.export_logs_message, fileName))
+        },
+
+        confirmButton = {
+            Button(
+                onClick = {
+
+
+                    coroutineScope.launch {
```
rajramsaroop, line 65: No need to wrap in coroutine scope as `exportLogs()` is not a suspending function (`exportLogs()` is using the `viewmodelScope` from the VM).

rajramsaroop (changes requested):

DittoExportLogs/src/main/java/live/ditto/dittoexportlogs/ExportLogs.kt
```diff
@@ -1,26 +1,97 @@
...
 import androidx.compose.ui.res.stringResource
-import live.ditto.exporter.ExportDialog
-import java.nio.file.Path
+import androidx.compose.ui.unit.dp
+import kotlinx.coroutines.launch
+import java.io.File
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
 
+@RequiresApi(Build.VERSION_CODES.O)
```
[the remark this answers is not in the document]
bplattenburg: @rajramsaroop Can you log a separate issue to follow up on this so it doesn't get lost?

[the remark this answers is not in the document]
rajramsaroop: Done: https://github.com/getditto/DittoAndroidTools/issues/141

DittoExportLogs/src/main/java/live/ditto/dittoexportlogs/ExportLogs.kt
zmarkan: good spot - I missed when moved it out into a ViewModel

zmarkan:
Implemented the requested changes

rajramsaroop (approved):

PR #145, opened 2024-12-13
update naming of peerKey to peerKeyString

texasRanger09:
closes CXTOOLS-203

The actual peerKey calls were already using peerKeyString. This PR simply updates the property name to peerKeyString


DittoHeartbeat/src/main/java/live.ditto.dittoheartbeat/Heartbeat.kt
```diff
@@ -94,7 +94,7 @@ fun addToCollection(info: DittoHeartbeatInfo, config: DittoHeartbeatConfig, ditt
         "metaData" to metaData,
         "sdk" to info.sdk,
         "_schema" to info.schema,
-        "peerKey" to info.peerKey
+        "peerKeyString" to info.peerKeyString
```
bplattenburg, line 97: I think the key name here actually needs to stay the same, Swift still has it as `peerKey` as far as I can tell.

bplattenburg (changes requested):

bplattenburg (approved):
Ship it!

PR #146, opened 2025-01-16
fix NPE that occurs on emulators without bluetooth, update tools version

rajramsaroop:
Fixes CXTOOLS-404

This fixes the above issue by checking if the bluetooth adapter returned is null. On an emulator without bluetooth this will show bluetooth as "not enabled", see screenshot:

Tapping "Enable Bluetooth" will take you to an Android settings screen, however the enable bluetooth toggle switch will be disabled. Going back into the app will crash the settings app, but there is nothing we can do about that because it's an OS app /shrug

I am not sure if that prevents the compatibility app from being published, I am now considering that if `null` is returned for the adapter, then to state that bluetooth feature doesn't exist (as opposed to "not enabled").


bplattenburg:
Would it be possible to show this state distinctly as `Unsupported` to mirror what we show on iOS simulators in DittoSwiftTools?

rajramsaroop:
> Would it be possible to show this state distinctly as `Unsupported` to mirror what we show on iOS simulators in DittoSwiftTools?

Yeah should be possible to do something like

true/false = enabled/disabled
null = unsupported

And update the UI accordingly.

rajramsaroop:
I've now made bluetooth a state as opposed to a boolean. So now it can either be `enabled`/`disabled`/`unsupported`. This is what unsupported looks like:



No change to what bluetooth enabled/disabled looks like.

bplattenburg (approved):

zmarkan:
I caused the conflicts so I can pick this up to get it mergeable ^^

rajramsaroop:
> I caused the conflicts so I can pick this up to get it mergeable ^^

I think I can do it! I'm aware of the changes so it shouldn't be that difficult!

bplattenburg (approved):

PR #147, opened 2025-01-20
Zan/unify packages remove modules

zmarkan:
Draft: Merging all tools into a single module - need to do that first before attempting any tree shaking. So far so good.


rajramsaroop:
I think we can go ahead and delete the old folders that used to be modules

README.md
rajramsaroop, line 1: I think we can either:

- get rid of the "Usage" section and move the information about the `local.properties` file into the "Testing in the demo app" section as that's where it's relevant
- change the "Usage" section to "View the tools in the demo app" type of section and just show how to run the demo app, in case someone just wants to view the tools in action vs. implementing it in their codebase

rajramsaroop (changes requested):

README.md
zmarkan: Good shout. I moved the local.properties bit down to the testing section. Usage now refers just to the actual usage of individual tools.

AndroidDittoTools/src/main/java/live/ditto/tools/health/data/HealthUiState.kt
```diff
@@ -1,13 +1,13 @@
...
 
 import android.content.Context
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
-import live.ditto.health.HealthUiActionType
-import live.ditto.health.HealthUiStateCause
-import live.ditto.health.R
-import live.ditto.health.usecase.WifiAwareState
+import live.ditto.tools.R.*
```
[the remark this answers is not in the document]
zmarkan: Good spot.

zmarkan:
Addressed all the above - ready to be reviewed again @rajramsaroop

rajramsaroop (approved):

PR #154, opened 2025-06-03
Ah/cxtools 158 peers list

al-ditto:
Fixes CXTOOLS-158

Add a new peers list view that is similar to the iOS peers list except the BLE distance portion of it since we are looking to remove BLE distance from the sdk. 

|Preview|
|------|
||


DittoToolsAndroid/src/main/java/live/ditto/tools/peerslist/PeersListViewer.kt
```diff
@@ -0,0 +1,138 @@
...
+@Composable
+fun PeersListViewer(
+    modifier: Modifier = Modifier,
+    ditto: Ditto
+) {
+    val peerListViewScope = rememberCoroutineScope()
+    val stateHolder = remember { PeerListViewStateHolder(ditto, peerListViewScope) }
+
+    Scaffold(
+        modifier = Modifier.fillMaxSize(),
```
rajramsaroop, line 46: Make this use the input `modifier`, i.e. it should be `modifier.fillMaxSize()`.

DittoToolsAndroid/src/main/java/live/ditto/tools/peerslist/PeersListViewer.kt
```diff
@@ -0,0 +1,138 @@
...
+                    style = MaterialTheme.typography.labelLarge
+                )
+            }
+            item {
+                PeerView(localPeer)
+            }
+
+            item {
+                Text(
+                    "REMOTE PEERS",
```
rajramsaroop, line 106: Extract text to string resource

DittoToolsAndroid/src/main/java/live/ditto/tools/peerslist/PeersListViewer.kt
```diff
@@ -0,0 +1,138 @@
...
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun PeersListViewer(
+    modifier: Modifier = Modifier,
+    ditto: Ditto
+) {
+    val peerListViewScope = rememberCoroutineScope()
+    val stateHolder = remember { PeerListViewStateHolder(ditto, peerListViewScope) }
+
+    Scaffold(
```
rajramsaroop, line 45: Hmm I'm actually thinking the Scaffold is too much UI. While it's nice that it gives us certain things like the top app bar, we want each individual tool to be made such that it can be included on it's own. 

Can we remove the Scaffold and incorporate the title and play/pause button into the Composable?

DittoToolsAndroid/src/main/java/live/ditto/tools/peerslist/PeerListViewStateHolder.kt
```diff
@@ -0,0 +1,42 @@
...
+import live.ditto.DittoPeer
+import androidx.compose.runtime.*
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.channels.awaitClose
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.callbackFlow
+import kotlinx.coroutines.launch
+import live.ditto.DittoPresenceGraph
+
+class PeerListViewStateHolder(val ditto: Ditto, scope: CoroutineScope) {
```
rajramsaroop, line 13: Conceptually having a class to hold the state is the right move, but I think this would be better off as a ViewModel. A class that implements `ViewModel` will also get us a `viewModelScope` among other benefits.

DittoToolsAndroid/src/main/java/live/ditto/tools/peerslist/PeerListViewStateHolder.kt
```diff
@@ -0,0 +1,42 @@
...
+    var remotePeers by mutableStateOf(listOf<DittoPeer>())
+        private set
+    var localPeer: DittoPeer = ditto.presence.graph.localPeer
+    var isPaused by mutableStateOf(false)
```
rajramsaroop, line 17: I would caution against using classes from a library as part of your state and instead create some sort of class to hold the elements that your UI cares about, maybe something like:

```
data class PeersListState(
val localPeer: Peer,
val remotePeers: List<Peer>,
val isPaused: Boolean)
```
You would then expose this in a view model typically using state flows. Take a look at `HealthViewModel` as an example

rajramsaroop:
Overall great implementation! Just a few adjustments to make it align more closely with Android best practices.

DittoToolsAndroid/src/main/java/live/ditto/tools/peerslist/PeersListViewer.kt
```diff
@@ -0,0 +1,138 @@
...
+                                    modifier = Modifier.size(24.dp)
+                                )
+                            }
+                        }
+                    }
+                )
+            }
+        }
+    ) { innerPadding ->
+        PeerListView(stateHolder.localPeer, stateHolder.remotePeers, padding = innerPadding)
```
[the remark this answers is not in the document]
al-ditto: Do you mean pass in the data object that holds both the localPeer and the remotePeers instead of separating them?

DittoToolsAndroid/src/main/java/live/ditto/tools/peerslist/PeerListViewStateHolder.kt
al-ditto: I moved the states to a data class object

DittoToolsAndroid/src/main/java/live/ditto/tools/peerslist/PeersListViewer.kt
al-ditto: Removed

DittoToolsAndroid/src/main/java/live/ditto/tools/peerslist/PeersListViewer.kt
```diff
@@ -0,0 +1,138 @@
...
+        LazyColumn(
+            modifier = Modifier
+                .fillMaxSize()
+                .padding(10.dp),
+            contentPadding = padding,
+            verticalArrangement = Arrangement.spacedBy(12.dp)
+        ) {
+            item {
+                Text(
+                    "LOCAL (SELF) PEER",
```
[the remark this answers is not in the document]
al-ditto: Extracted

DittoToolsAndroid/src/main/java/live/ditto/tools/peerslist/PeersListViewer.kt
```diff
@@ -0,0 +1,138 @@
...
+    val peerListViewScope = rememberCoroutineScope()
+    val stateHolder = remember { PeerListViewStateHolder(ditto, peerListViewScope) }
+
+    Scaffold(
+        modifier = Modifier.fillMaxSize(),
+        topBar = {
+            Surface(tonalElevation = 4.dp) {
+                TopAppBar(
+                    title = {
+                        Text("Peers List")
```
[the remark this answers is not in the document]
al-ditto: Extracted

al-ditto:
> Overall great implementation! Just a few adjustments to make it align more closely with Android best practices.

I noticed an issue which I did not dive deep into but the DittoTools app crashes when I change orientation twice. Logcat indicates the app attempts to create the Ditto instance multiple times

DittoToolsAndroid/src/main/java/live/ditto/tools/peerslist/PeersListViewer.kt
```diff
@@ -0,0 +1,138 @@
...
+        shape = RoundedCornerShape(12.dp),
+        colors = CardDefaults.cardColors(
+            containerColor = MaterialTheme.colorScheme.secondaryContainer
+        )
+    ) {
+        Box(modifier = Modifier.padding(16.dp)) {
+            Column {
+                Text("${peer.deviceName}:", style = MaterialTheme.typography.titleLarge)
+                Text(peer.peerKeyString)
+                peer.dittoSdkVersion?.let { Text("SDK v${it}") }
```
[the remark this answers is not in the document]
rajramsaroop: Haha I was meant to type out that when you hover over it, the yellow light bulb appears and when you click on it, there's an option to "extract to string resource". It was early in the morning/I was in a rush to get to my doc appointment, but it looks like you figured it out!

DittoToolsAndroid/src/main/java/live/ditto/tools/peerslist/PeersListViewer.kt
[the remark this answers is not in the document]
rajramsaroop: I meant making use of the named arguments, see here for Kotlin documentation that would explain it better than me: https://kotlinlang.org/docs/functions.html#named-arguments

rajramsaroop (approved):

PR #159, opened 2025-10-29
Update navigation and ToolsMenu UI to match DittoSwiftTools

acoyner:
- Replaced bottom bar with back and exit buttons in TopBar
- Updated menu design to match DittoSwiftTools with sections (Network, System, Data & Debugging) and icons for all menu items
- TopAppBar reads and matches the parent app's status bar color
- Updated app/build.gradle to use local project module (project(':DittoToolsAndroid')) instead of published package for testing

Before:
Screenshot 2025-10-29 at 11 30 30 AM

After:
Screenshot 2025-10-29 at 11 40 08 AM


DittoToolsAndroid/src/main/java/live/ditto/tools/toolsviewer/DittoToolsViewer.kt
```diff
@@ -69,28 +74,96 @@ private fun DittoToolsViewerScaffold(
...
+    val isSystemInDarkTheme = androidx.compose.foundation.isSystemInDarkTheme()
+    val backgroundColor = if (isSystemInDarkTheme) Color(0xFF1C1C1E) else Color(0xFFF2F2F7)
+
+    // Handle system back button
+    BackHandler(enabled = true) {
+        if (isMainScreen) {
+            onExitTools()
+        } else {
+            navController.popBackStack()
+        }
```
rajramsaroop: Can probably put this in a function since its repeated when you click the back arrow

DittoToolsAndroid/src/main/java/live/ditto/tools/toolsviewer/viewmodel/ToolsViewerViewModel.kt
```diff
@@ -1,50 +1,85 @@
...
 class ToolsViewerViewModel: ViewModel() {
 
-    fun toolsMenuItems(): List<ToolMenuItem> {
+    fun toolsMenuSections(): List<ToolMenuSection> {
         return listOf(
-            ToolMenuItem(
-                label = R.string.presence_viewer_tool_label,
-                route = Screens.PresenceViewerScreen.route
+            ToolMenuSection(
+                title = "Network",
```
rajramsaroop: lets put user facing string in string resources

acoyner: Done

DittoToolsAndroid/src/main/java/live/ditto/tools/toolsviewer/Color.kt
rajramsaroop: This should be in ui/theme package with other theme related files

DittoToolsAndroid/src/main/java/live/ditto/tools/toolsviewer/DittoToolsViewer.kt
```diff
@@ -69,28 +74,92 @@ private fun DittoToolsViewerScaffold(
...
+    val navBackStackEntry by navController.currentBackStackEntryAsState()
+    val currentRoute = navBackStackEntry?.destination?.route
+    val isMainScreen = currentRoute == Screens.MainScreen.route
+
+    // Read parent app's status bar color
+    val view = LocalView.current
+    val statusBarColor = remember {
+        if (!view.isInEditMode) {
+            val colorInt = (view.context as? android.app.Activity)?.window?.statusBarColor
+            colorInt?.let { Color(it) } ?: Color(0xFF6200EE) // Fallback to material purple
```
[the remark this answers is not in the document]
rajramsaroop: The fallback color can be added to Colors.kt

DittoToolsAndroid/build.gradle.kts
```diff
@@ -27,6 +27,7 @@ dependencies {
     implementation(libs.androidx.lifecycle.runtime.compose)
     implementation(platform(libs.androidx.compose.composeBom))
     implementation(libs.androidx.compose.runtime.runtimeLivedata)
+    implementation(libs.androidx.compose.material.icons.extended)
```
rajramsaroop: Can you create a linear ticket to bundle the material icons with tools, as opposed to including the library?

Out of scope for this PR.

But including the material icons as a dependency increases apk size, since its including all icons (and variants). We should be mindful to not increase a customer's apk size if not needed. 

I am not sure if gradle would be smart enough to not include these resources if not needed at compile time, that might be something, as part of the ticket you create, to see if the apk size is significantly reduced by removing this library/dependency and just including the icons we want

rajramsaroop (approved):
One comment regarding creating a Linear ticket, but doesn't need to be a blocker for approving, so I'm hitting approve 👍🏾

PR #161, opened 2025-10-30
Update to 4.11.6

acoyner:
- Update Android Tools to 4.11.6
- Update demo app to include authUrl and websocketUrl, disableSyncWithV3, and strict mode = false

Implements CXTOOLS-565

### DIFF
--- a/app/src/main/java/live/ditto/dittotoolsapp/MainActivity.kt
+++ b/app/src/main/java/live/ditto/dittotoolsapp/MainActivity.kt
@@ -51,7 +51,6 @@ class MainActivity : ComponentActivity() {
                     LaunchedEffect(key1 = true) {
                         try {
                             ditto = createDitto()
-                            ditto?.startSync()
                         } catch (e: Throwable) {
                             dittoError = e.message.toString()
                             Log.e("Ditto error", e.message.toString())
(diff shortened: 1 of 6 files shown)


app/src/main/java/live/ditto/dittotoolsapp/MainActivity.kt
```diff
@@ -85,17 +84,32 @@ class MainActivity : ComponentActivity() {
...
+            dependencies = androidDependencies,
             appId = BuildConfig.DITTO_ONLINE_PLAYGROUND_APP_ID,
             token = BuildConfig.DITTO_ONLINE_PLAYGROUND_TOKEN,
-            enableDittoCloudSync = true
+            customAuthUrl = BuildConfig.DITTO_CUSTOM_AUTH_URL,
+            enableDittoCloudSync = false
         )
-        val ditto = Ditto(androidDependencies, identity)
-        DittoLogger.minimumLogLevel = DittoLogLevel.DEBUG
+        val ditto = Ditto(
```
rajramsaroop: In an ideal world the instance of Ditto is held in a singleton instance class (like a DittoManager), typically injected so that it survives android configuration changes.

Since this is a basic demo app, I'd say we can move the ditto instance, and related creation functions to an `Application` class, which would give us the same effect

acoyner: Done

acoyner: Are you setting the customauthurl and websocketurl? It's not getting stuck on mine

rajramsaroop: Once its in the application class, you should be able to rotate the device without the Ditto instance trying to recreate itself (which was one of the main issues of having it in the MainActivity)

rajramsaroop (approved):

PR #162, opened 2025-10-30
Implements Export Callback for DiskUsage tool

acoyner:
Adds an optional onExport callback parameter to DittoDiskUsage to allow custom export logic. When a callback is provided, the Android file picker is unused and the exported zip is passed directly to the callback. If no callback is provided, the default file picker will be used

Implements CXTOOLS-544

Demo:
https://www.loom.com/share/806c8b2b155c42a1806c5fbbb75dfdac

### DIFF
--- a/DittoToolsAndroid/src/main/java/live/ditto/tools/diskusage/DiskUsage.kt
+++ b/DittoToolsAndroid/src/main/java/live/ditto/tools/diskusage/DiskUsage.kt
@@ -9,10 +9,14 @@ import androidx.lifecycle.viewmodel.compose.viewModel
  * Wrapper composable function for `DiskUsageView`.
  */
 @Composable
-fun DiskUsageScreen() {
+fun DiskUsageScreen(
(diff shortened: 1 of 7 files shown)


DittoToolsAndroid/src/main/java/live/ditto/tools/diskusage/DiskUsage.kt
```diff
@@ -9,10 +9,14 @@ import androidx.lifecycle.viewmodel.compose.viewModel
  * Wrapper composable function for `DiskUsageView`.
  */
 @Composable
-fun DiskUsageScreen() {
+fun DiskUsageScreen(
+    onExport: ((java.io.File) -> Unit)? = null
```
rajramsaroop: you can add an import for java.io.File so you can have `(File) -> Unit`

rajramsaroop:
Lets update the readme with how to use as well

rajramsaroop (approved):

PR #163, opened 2025-11-04
feat: Add note about DQL_STRICT_MODE for Export Logs

mrjeremystookey:
Add a note to the README clarifying that `DQL_STRICT_MODE` must be `false` for the "Export Logs to Portal" feature to work.


bplattenburg (approved):
Thanks!

PR #165, opened 2025-11-10
Downgraded dependencies and related code to SDK 33, Kotlin < 2

acoyner:
- Downgraded compileSdk and targetSdk from 34 to 33, along with Compose BOM (2023.06.01) and Material library (1.10.0) to avoid API 34-only resources
  - Downgraded Kotlin from 2.1.0 to 1.9.25 with Compose compiler 1.5.15 for compatibility with Kotlin 1.x consumers
  - Added dependency resolution strategies to force SDK 33-compatible versions of androidx.activity (1.7.2), material3 (1.0.1), and material (1.10.0)
  - Updated Compose APIs: replaced HorizontalDivider with Divider and Icons.AutoMirrored.* with Icons.Filled.* for Compose 1.4.x compatibility
  
  Fixes CXTOOLS-212

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,6 @@ plugins {
     alias libs.plugins.com.android.application apply false
     alias libs.plugins.com.android.library apply false
     alias libs.plugins.org.jetbrains.kotlin.android apply false
-    alias(libs.plugins.compose.compiler) apply false
     alias(libs.plugins.jreleaser) apply false
 }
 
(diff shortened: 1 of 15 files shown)


build.gradle
```diff
@@ -27,6 +26,28 @@ ext.getValueFromLocalProperties = { name, defValue ->
     return (null == prop) ? defValue : prop
 }
 
+subprojects {
```
rajramsaroop: Can you add a comment here to say this is enforcing dependency constraints only for subprojects in Ditto Tools, and that its not enforced on consuming projects?

rajramsaroop:
Were we able to test this on an external app that had these issues?

build.gradle
acoyner: Done

rajramsaroop:
> @rajramsaroop Do we have any demo apps that have these issues?

Not sure, don't think so. I tried creating an empty android project last night and setting the compile sdk to 33, but ran into some compile issues.

If you have time, you could try testing that way, or check if any of our demo apps or quickstart apps use compile sdk 33 or lower.

rajramsaroop (approved):
Nice work, this unblocks a few customers, thank you!

PR #166, opened 2026-01-06
Convert experimental text field to stable basic text field

chrisroberts14:
Related to Zendesk Ticket [2667](https://dittosync.zendesk.com/agent/tickets/2667).
When opening the Presence Degradation Reporter an app integrating the tools would crash (but the tools alone would not). 
This could be reproduced by putting the tools into the QS app. 
My theory is that since the `TextField` component is marked as experimental in the version of material3 we are using the API changed in later versions and apps could implicitly upgrade it causing the crash.

We can't upgrade as we are maintaining compatibility with API v33 and forcing users to use an older version of material3 is not desirable.

My solution is to change the page slightly to use the the stable `BasicTextField` this stops the crash.
| Before    | After |
| -------- | ------- |
|  image | image|


zmarkan (approved):
That's sensible. Experimental can result in differing underlying implementations across version - which explains why our demo app works and theirs doesn't.

PR #170, opened 2026-03-06
Migrate DittoAndroidTools to Ditto Kotlin SDK v5

kristopherjohnson:
## Summary

Migrate DittoAndroidTools library from Ditto SDK v4 (`live.ditto:ditto:4.11.6`) to v5 (`com.ditto:ditto-kotlin:5.0.0-rc.1`). All tools remain functionally and visually identical to v4 except where noted.

### Changes across all tools
- Package imports: `live.ditto.*` → `com.ditto.kotlin.*`
- Property renames: `peerKeyString` → `peerKey`, `isConnectedToDittoCloud` → `isConnectedToDittoServer`, `sdkVersion` → `Ditto.VERSION`, `persistenceDirectory` → `absolutePersistenceDirectory`
- `DittoError` → `DittoException` (caught as `Exception`)
- Presence observation: v4 callback-based `observe {}` wrapped in `callbackFlow` → v5 `observe()` returns `Flow` directly (deleted manual `callbackFlow` wrappers in Presence Viewer, Peers List, and Presence Degradation Reporter)

### Data Browser
- Collections discovery: `store.collections().observeLocal {}` → `store.registerObserver("SELECT name FROM __collections")` with reactive updates (no polling)
- Standalone mode syncs `__collections` via `sync.registerSubscription("SELECT * FROM __collections")`
- Document queries: `collection.findAll().observeLocal {}` → `store.registerObserver("SELECT * FROM ...")`
- Document values: direct `doc.value` map → CBOR deserialization via `DittoCborSerializable` with null-safe type coercion; recursively unwraps `Dictionary` and `ArrayValue` types into readable maps/lists
- Subscriptions: removed redundant per-document subscription — `CollectionsViewModel` already handles standalone subscriptions at the collection level
- Stand-alone mode: restored v4 behavior where "Start Subscriptions" immediately syncs all known collections (and picks up newly discovered ones)
- Added `onCleared()` lifecycle cleanup for observers and subscriptions (missing in v4)
- Dark theme: explicit `MaterialTheme.colorScheme.onBackground` for text colors
- All DQL collection name interpolations backtick-quoted for safety with non-standard names

### Disk Usage
- Observation: callback-based `diskUsage.observe {}` with `DisposableEffect` → Flow-based `diskUsage.observe().collect {}` in `LaunchedEffect`
- `DiskUsageItem` → `DittoDiskUsageItem`
- `sizeInBytes` type: `Int` → `Long` throughout the disk usage chain
- Strip persistence directory prefix from paths (v5 appends database ID to persistence directory)

### Heartbeat
- Store write: `store.collection(name).upsert(doc)` → `store.execute("INSERT INTO ... ON ID CONFLICT DO UPDATE")` with explicit CBOR serialization via `toDittoCbor()`
- `addToCollection` became `suspend` (v5 `store.execute()` is suspending)

### Presence Degradation Reporter
- Store operations: `store[collection].upsert()`, `findAll().remove()`, `transaction.scoped()` → DQL (`INSERT ... ON ID CONFLICT DO UPDATE`, `EVICT FROM ... WHERE true`)
- Reactive queries: `observeLocalAsFlow()` custom extension → `registerObserver()` with `MutableStateFlow`
- Model deserialization: `DittoDocument.stringValue` → `DittoCborSerializable.Dictionary` with null-safe defaults
- Added `close()` method for observer cleanup
- Deleted `ditto/DittoExt.kt` (`DittoPendingCursorOperation.observeLocalAsFlow()` no longer needed)

### Log Viewer
- Log directory path: `filesDir/ditto/ditto_logs` → `ditto.absolutePersistenceDirectory/ditto_logs` (v5 appends database ID)
- Removed `filesDir` parameter from `LogUtils`, ViewModels, factories, and Compose screens

### Export Logs
- Confirmation text shows `peerKey` instead of `appId` (removed in v5)

### Tools Viewer
- Replaced deprecated `Window.statusBarColor` with `MaterialTheme.colorScheme.primary`

### Presence Viewer
- Added `THIRD_PARTY_LICENSES.txt` for bundled JS dependencies (vis-network, vis-data, Hammer.js, keycharm, color, core-js, uuid)

### No changes (import-only)
- Peers List, Health Viewer, Exporter

## Test plan
- [x] `./gradlew assembleDebug` builds clean
- [x] `./gradlew :DittoToolsAndroid:assembleRelease` builds clean
- [x] Demo app launches and tools are functional on device (Galaxy S20 FE, Pixel 3)
- [x] Verify Data Browser collection listing and document display
- [x] Verify Presence Viewer mesh graph renders
- [x] Verify Disk Usage metrics display
- [x] Verify Log Viewer and Log Details screens
- [ ] Verify Export Logs to Portal

**Linear:** [SDKS-1489](https://linear.app/ditto/issue/SDKS-1489/update-dittoandroidtools-to-use-ditto-kotlin-sdk-v5)


app/src/main/java/live/ditto/dittotoolsapp/DittoToolsApplication.kt
```diff
@@ -25,30 +26,23 @@ class DittoToolsApplication : Application() {
...
-                    updateTransportConfig { transportConfig ->
-                        transportConfig.connect.websocketUrls.add(BuildConfig.DITTO_WEBSOCKET_URL)
+                    auth?.expirationHandler = { dittoInstance, _ ->
+                        dittoInstance.auth?.login(
+                            token = BuildConfig.DITTO_ONLINE_PLAYGROUND_TOKEN,
+                            provider = DittoAuthenticationProvider.development(),
+                        )
                     }
-
                     store.execute("ALTER SYSTEM SET DQL_STRICT_MODE = false")
```
chrisroberts14: Is this still necessary in v5?

kristopherjohnson: No, it's not. Removed.

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/DocumentsViewModel.kt
```diff
@@ -16,54 +19,60 @@ class DocumentsViewModel(private val collectionName: String, isStandAlone: Boole
...
-        applyFilter()
+    val subscription: DittoSyncSubscription? = if (isStandAlone) {
+        DittoHandler.ditto.sync.registerSubscription("SELECT * FROM $collectionName LIMIT 1000")
+    } else {
+        null
     }
 
-    private fun findAllLiveQuery() {
-        this.liveQuery =  DittoHandler.ditto.store.collection(collectionName).findAll().limit(1000).observeLocal { docs, _ ->
+    private var liveQuery: DittoStoreObserver = createFindAllObserver()
```
chrisroberts14: This is slightly nit-picky but we have previously tried to move away from the live query terminology in places customers might see it (see SPO-4 on linear). Since this code is public ideally we would also use observer terminology here.

DittoToolsAndroid/src/main/java/live/ditto/tools/heartbeat/Heartbeat.kt
```diff
@@ -34,13 +34,13 @@ data class DittoHeartbeatInfo(
...
      */
-    var healthMetrics: MutableMap<String, HealthMetric> = mutableMapOf()
+    var healthMetrics: MutableMap<String, live.ditto.tools.healthmetrics.HealthMetric> = mutableMapOf()
 )
 
 lateinit var info: DittoHeartbeatInfo
 
 private fun updateHealthMetrics(config: DittoHeartbeatConfig) {
-    val newHealthMetrics = mutableMapOf<String, HealthMetric>()
+    val newHealthMetrics = mutableMapOf<String, live.ditto.tools.healthmetrics.HealthMetric>()
```
chrisroberts14: Is there a specific reason to not import the `HealthMetric` so these types are shorter?

DittoToolsAndroid/src/main/java/live/ditto/tools/toolsviewer/DittoToolsViewer.kt
chrisroberts14: This seemingly no longer works on Android 15+ it just results in a white banner.
Claude gave me the suggestion:
```
 // Replace this (lines 90–99):
  val view = LocalView.current
  val statusBarColor = remember {
      if (!view.isInEditMode) {
          val colorInt = (view.context as? android.app.Activity)?.window?.statusBarColor
          colorInt?.let { Color(it) } ?: Color(0xFF6200EE)
      } else {
          Color(0xFF6200EE)
      }
  }

  // With this:
  val statusBarColor = MaterialTheme.colorScheme.primary
```
Which seemingly works.

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/CollectionsViewModel.kt
```diff
@@ -2,23 +2,64 @@ package live.ditto.tools.databrowser
...
     var isStandAlone = false
 
-    private lateinit var subscription: DittoSubscription
-    val liveQuery = DittoHandler.ditto.store.collections().observeLocal { collections ->
-        this.collections.postValue(collections.collections)
+    private val subscriptions = mutableMapOf<String, DittoSyncSubscription>()
+
+    // system:collections is a virtual collection that doesn't support registerObserver,
+    // so we poll periodically with store.execute().
+    private val pollJob = viewModelScope.launch(Dispatchers.IO) {
```
chrisroberts14: For use as a standalone app we need to subscribe to the `__collections` (I don't think it works with `system:collections`) collection to actually get all the collection names. But we can register an observer against `__collections`.

My suggestion would be replacing the `startSubscription` function with just starting a subscription to `__collections` and you can just have an observer instead of polling.

You might need to figure out how to close the subscription and observer when the view model is closed.
```kotlin
val observer = DittoHandler.ditto.store.registerObserver(
    "SELECT name FROM __collections"
) {
    val names = it.items.mapNotNull { item ->
        item.value["name"].stringOrNull
    }
    collections.postValue(names)
    subscribeToCollections(names)
}

fun startSubscription() {
    DittoHandler.ditto.sync.registerSubscription(
        "SELECT * FROM __collections"
    )
    // Also run a query to populate the initial state
    viewModelScope.launch(Dispatchers.IO) {
        val result = DittoHandler.ditto.store.execute("SELECT name FROM __collections")
        val names = result.items.mapNotNull { item ->
            item.value["name"].stringOrNull
        }
        collections.postValue(names)
        subscribeToCollections(names)
    }
}
```

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/CollectionsViewModel.kt
```diff
@@ -2,23 +2,64 @@ package live.ditto.tools.databrowser
...
         isStandAlone = true
+        // Subscribe to all currently known collections
+        collections.value?.let { subscribeToCollections(it) }
     }
 
+    private fun subscribeToCollections(names: List<String>) {
+        for (name in names) {
+            if (name !in subscriptions) {
+                subscriptions[name] = DittoHandler.ditto.sync.registerSubscription(
+                    "SELECT * FROM $name"
```
chrisroberts14: You'll need to quote the collection name as this will fail if `name` begins with a number as that doesn't work in DQL
```suggestion
                    "SELECT * FROM `$name`"
```

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/Collections.kt
chrisroberts14: This top bar seems to somehow throw the whole UI on this page off. Given we already have "Data Browser" in the nav bar at the top I'd say we can just remove the topBar entirely.
It'll end up looking like 

Image

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/Documents.kt
```diff
@@ -180,13 +181,17 @@ fun DocItem(property: String, viewModel: DocumentsViewModel, selectedDoc: Docume
...
+            color = MaterialTheme.colorScheme.onBackground,
             modifier = Modifier.weight(1f)
         )
         val doc: Document? = viewModel.docsList.value?.find {
             it == selectedDoc
         }
         if (doc != null) {
-            Text(doc.properties[property].toString())
+            Text(
+                text = doc.properties[property].toString(),
```
chrisroberts14: This seems to give a quite user unfriendly result when it comes to arrays and json objects. It might be the case we need specific handling for arrays and dictionaries in `cborToDisplayValue`
Before:

Image

Now:

Image

chrisroberts14 (changes requested):
Apologies for the disjointed reviews I've now managed to pull down the repo and I've found some issues. I haven't been able to look through everything though.

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/DocumentsViewModel.kt
```diff
@@ -114,8 +120,23 @@ class DocumentsViewModel(private val collectionName: String, isStandAlone: Boole
...
+    private fun cborToDisplayValue(cbor: DittoCborSerializable): Any? {
+        return cbor.stringOrNull
+            ?: cbor.longOrNull
+            ?: cbor.booleanOrNull
+            ?: cbor.doubleOrNull
+            ?: cbor.floatOrNull
+            ?: if (cbor.isNull) null else cbor.toString()
+    }
+
     class MyViewModelFactory(private val collectionName: String, private val isStandAlone: Boolean) :
```
chrisroberts14: Not something you did but this should be changed to match the name of the view model, not the most important thing though

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/DocumentsViewModel.kt
```diff
@@ -16,54 +19,60 @@ class DocumentsViewModel(private val collectionName: String, isStandAlone: Boole
...
+        observer = createFindAllObserver()
+    }
+
+    private fun findWithFilterObserver(queryString: String) {
         try {
             errorMessage.postValue(null)
-            this.liveQuery =  DittoHandler.ditto.store.collection(collectionName).find(queryString).limit(1000).observeLocal { docs, _ ->
+            observer.close()
+            observer = DittoHandler.ditto.store.registerObserver(
+                "SELECT * FROM $collectionName WHERE $queryString LIMIT 1000"
```
chrisroberts14: The `collectionName` should be quoted here. I like using backticks (`) for this since they are so rarely used as quotes in strings.

```suggestion
                "SELECT * FROM `$collectionName` WHERE $queryString LIMIT 1000"
```

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/DocumentsViewModel.kt
```diff
@@ -16,54 +19,60 @@ class DocumentsViewModel(private val collectionName: String, isStandAlone: Boole
...
+        null
     }
 
-    private fun findAllLiveQuery() {
-        this.liveQuery =  DittoHandler.ditto.store.collection(collectionName).findAll().limit(1000).observeLocal { docs, _ ->
+    private var observer: DittoStoreObserver = createFindAllObserver()
+
+    private fun createFindAllObserver(): DittoStoreObserver {
+        return DittoHandler.ditto.store.registerObserver(
+            "SELECT * FROM $collectionName LIMIT 1000"
```
chrisroberts14: Also would need to be quoted.

```suggestion
            "SELECT * FROM `$collectionName` LIMIT 1000"
```

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/DocumentsViewModel.kt
```diff
@@ -16,54 +19,60 @@ class DocumentsViewModel(private val collectionName: String, isStandAlone: Boole
...
-            }
-            newDocsList.add(Document(doc.id.toString(), docValues))
-        }
-        allDocuments = newDocsList
-        applyFilter()
+    val subscription: DittoSyncSubscription? = if (isStandAlone) {
+        DittoHandler.ditto.sync.registerSubscription("SELECT * FROM $collectionName LIMIT 1000")
+    } else {
+        null
     }
```
chrisroberts14: This subscription is surplus and should be removed entirely (along with the `isStandAlone` variable in this view model). In the case of it being a standalone app we already subscribe to the collection when the user presses the button on the data browser screen

DittoToolsAndroid/src/main/java/live/ditto/tools/diskusage/DiskUsageViewModel.kt
chrisroberts14: Since in v5 we now append the database id to the persistence directory this path causes the screen to no longer show the sizes. The easiest way would be to remove the persistence directory name and only show the files/directory names in it.  Though there is probably a more robust solution.

```suggestion
                relativePath = record.path.split("/")[1]
```

Before (v4 version):
Image
After (Your branch with the above change made):
Image

DittoToolsAndroid/src/main/java/live/ditto/tools/utils/LogUtils.kt
chrisroberts14: Since we now append the database Id to the persistence directory this is no longer correct but we can easily use the absolute persistence directory here.
```suggestion
    private val dittoLogDir = File("${ditto.absolutePersistenceDirectory}/ditto_logs")
```
This also means we can remove the `filesDir` argument.

chrisroberts14 (changes requested):
Just a few more issues I found when running through the whole toolset.

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/CollectionsViewModel.kt
kristopherjohnson: I'm not sure we can subscribe to a virtual collection like `__collections`. But I'll give it a try.

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/Collections.kt
kristopherjohnson: I agree the UI needs an overhaul. The close box in the upper right does nothing on any of the screens. But that's outside the scope of this PR. I'll create another issue for UI improvements.

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/CollectionsViewModel.kt
chrisroberts14: I did give this a go when I was running through and it worked but I know it doesn't when you use `system:...`

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/Collections.kt
chrisroberts14: Agreed for now I think it should be fine to remove the topbar

kristopherjohnson:
I've created this issue as a holding place for suggested changes that are outside the scope of this PR's focus (migration to the v5 SDK): https://linear.app/ditto/issue/CXTOOLS-645/dittoandroidtools-usability-improvements

DittoToolsAndroid/src/main/java/live/ditto/tools/toolsviewer/DittoToolsViewer.kt
kristopherjohnson: Fixed

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/CollectionsViewModel.kt
kristopherjohnson: I started a thread about `__collections` vs `system:collections` and observability here: https://dittolive.slack.com/archives/C04DUB2JP6C/p1773422363899899

Will address this when the answers are clear.

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/CollectionsViewModel.kt
kristopherjohnson: Fixed.

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/Collections.kt
kristopherjohnson: I've created a separate issue for this: https://linear.app/ditto/issue/CXTOOLS-648/dittoandroidtools-remove-or-fix-top-bar

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/DocumentsViewModel.kt
kristopherjohnson: Fixed

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/DocumentsViewModel.kt
kristopherjohnson: Fixed

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/Documents.kt
kristopherjohnson: Fixed

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/DocumentsViewModel.kt
kristopherjohnson: Fixed. Renamed to `DocumentsViewModel.Factory`.

DittoToolsAndroid/src/main/java/live/ditto/tools/diskusage/DiskUsageViewModel.kt
kristopherjohnson: Fixed. Using `record.path.substringAfter("/")`, which will work properly even if there is no `/` in the path.

kristopherjohnson:
### Code review

Found 4 issues:

1. **Resource leak: `PeersRepository` observers never closed in `PresenceDegradationReporterViewModel`** (CLAUDE.md says "Always close `DittoStoreObserver` in `onCleared()`. v4 code often leaked observers.")

`PeersRepository` holds 3 `DittoStoreObserver` instances and exposes `close()`, but the ViewModel never overrides `onCleared()` to call `repository.close()`. All three observers leak when the ViewModel is destroyed.

https://github.com/getditto/DittoAndroidTools/blob/a3fee0982a05812157c848cdc8627598f46a72f5/DittoToolsAndroid/src/main/java/live/ditto/tools/presencedegradationreporter/PresenceDegradationReporterViewModel.kt#L22-L27

2. **Resource leak: `PeersRepository` observers never closed in `Api.kt`** (CLAUDE.md says "Always close `DittoStoreObserver` in `onCleared()`")

`presenceDegradationReporterFlow()` creates a `PeersRepository` but never calls `repository.close()` when the Flow completes or is cancelled. The 3 store observers leak. Fix: add `.onCompletion { repository.close() }` before `.filterNotNull()`.

https://github.com/getditto/DittoAndroidTools/blob/a3fee0982a05812157c848cdc8627598f46a72f5/DittoToolsAndroid/src/main/java/live/ditto/tools/presencedegradationreporter/Api.kt#L21-L24

3. **Public API field `peerKeyString` not renamed to `peerKey`** (CLAUDE.md v5 migration reference says "peerKeyString → peerKey")

`DittoHeartbeatInfo` data class retains the v4 field name `peerKeyString`. The underlying SDK call was correctly updated to `ditto.presence.graph.localPeer.peerKey`, but the public data class field wasn't renamed.

https://github.com/getditto/DittoAndroidTools/blob/a3fee0982a05812157c848cdc8627598f46a72f5/DittoToolsAndroid/src/main/java/live/ditto/tools/heartbeat/Heartbeat.kt#L34-L36

4. **Trailing whitespace on modified line** (CLAUDE.md says "Trim trailing whitespace and end files with a blank line")

Line 42 contains a trailing space character. This file was modified by the PR.

https://github.com/getditto/DittoAndroidTools/blob/a3fee0982a05812157c848cdc8627598f46a72f5/DittoToolsAndroid/src/main/java/live/ditto/tools/logviewer/LogDetailsScreen.kt#L41-L43

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

kristopherjohnson:
I noticed this issue while looking at the Peers List: https://linear.app/ditto/issue/SDKS-2993/v4-peers-do-not-report-v5-sdk-version

This seems to be a Ditto core issue, rather than a problem with DittoAndroidTools. I don't think it is in the scope of this PR.

DittoToolsAndroid/src/main/java/live/ditto/tools/databrowser/CollectionsViewModel.kt
kristopherjohnson: OK, I've done observer and subscription on the `__collections` collection. Behavior seems to match the old DittoAndroidTools.

DittoToolsAndroid/src/main/java/live/ditto/tools/utils/LogUtils.kt
kristopherjohnson: Fixed.

chrisroberts14:
I've done a run through the tools and everything Ditto related seems to work fine. There are still a few UI issues but if we are considering those out of the scope of this PR I am happy to approve it.

kristopherjohnson:
@chrisroberts14 I think everything else is out of scope. The main point is to get this compatible with v5, and I'd like to close it out for last sprint if we've hit that goal.

chrisroberts14 (approved):

kristopherjohnson:
I'm going to leave this unmerged until v5 is released.

kristopherjohnson:
Merging this into the `sdk-5.0` branch, which will be merged into `main` when v5 is released.

PR #172, opened 2026-03-18
Handle unknown SDK version gracefully in peers list and heartbeat

alycda:
## Summary
Improved handling of unknown SDK versions across the peers list UI and heartbeat telemetry by adding explicit null checks and fallback values.

## Key Changes
- **PeersListViewer.kt**: Replaced the concise `?.let` expression with an explicit null check that displays "SDK version unknown" when `dittoSdkVersion` is null, providing better UX feedback
- **Heartbeat.kt**: Updated the connection map to use the peer's SDK version when available, falling back to the local Ditto SDK version as a default
- **strings.xml**: Added new string resource `sdk_version_unknown` for displaying when a peer's SDK version information is unavailable

## Implementation Details
The changes ensure that:
1. Users see a clear message when a remote peer's SDK version cannot be determined, rather than having the field omitted
2. Heartbeat telemetry includes the most accurate SDK version information available (peer-specific when present, otherwise local)
3. The fallback behavior maintains backward compatibility while improving data completeness

https://claude.ai/code/session_01BcPUrCDTy3kUi5QhxaZkbN

### DIFF
--- a/DittoToolsAndroid/src/main/java/live/ditto/tools/heartbeat/Heartbeat.kt
+++ b/DittoToolsAndroid/src/main/java/live/ditto/tools/heartbeat/Heartbeat.kt
@@ -113,7 +113,7 @@ fun getConnections(
 
         val connectionMap: Map<String, Any?> = mapOf(
             "deviceName" to connection.deviceName,
-            "sdk" to ditto.sdkVersion,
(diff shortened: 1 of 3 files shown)


alycda:
related: https://github.com/getditto/ditto/pull/21531

